PR #2, opened 2020-01-20
Setup PlatformIO project for Arduino firmware development

### LINKED ISSUE #1: Setup PlatformIO project for Arduino firmware development

alexproskurin:
Resolves #1


firmware/.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,63 @@
+{
```
noaht11: This file is full of paths specific to your computer. We should either gitignore it (if it can be automatically re-generated by platformio) or find a way so it doesn't use absolute paths.

Can you test what happens if you close the project, delete this file, then re-open it

firmware/.vscode/launch.json
```diff
@@ -0,0 +1,34 @@
+// AUTOMATICALLY GENERATED FILE. PLEASE DO NOT MODIFY IT MANUALLY
```
noaht11: This file is full of paths specific to your computer. We should either gitignore it (if it can be automatically re-generated by platformio) or find a way so it doesn't use absolute paths.

Can you test what happens if you close the project, delete this file, then re-open it

noaht11 (changes requested):
The commit messages don't seem accurate...I don't see an initial stepper motor class definition, but that's okay cause it probably shouldn't be part of this PR anyway. (Just an observation).

We do need to fix the user-specific path issues though

firmware/lib/StepperMotor/src/StepperMotor.hpp
```diff
@@ -0,0 +1,57 @@
...
+ * 
+ * Pin5 and 6 are connected to the other motor driver with same schematic.
+ * 
+ * Limit switch is meant to connect to pin4, directly from ground with no resistor is fine, current is divided 
+ * so no load is nessaccary. This part of code is commented out but can be added just by uncommenting.
+ */
+
+// =======================================
+// include guard
+#ifdef __STEPPERMOTOR_H_INCLUDED__
```
noaht11: Switch to `_STEPPERMOTOR_H_`

noaht11 (changes requested):

noaht11:
Abandoning in favour of #3

PR #5, opened 2020-02-24
Stepper motor code rework

alexproskurin:


firmware/src/main.cpp
```diff
@@ -1,16 +1,32 @@
...
 
+
 void loop()
 {
-    // put your main code here, to run repeatedly:
-    delay(1000);
-    digitalWrite(LED_BUILTIN, HIGH);
-    delay(1000);
-    digitalWrite(LED_BUILTIN, LOW);
-}
```
noaht11, line 16: Leave this in for now I think, it's a convenient way to know the code is running

firmware/lib/Axis/src/Axis.h
```diff
@@ -0,0 +1,63 @@
...
+// isr for operating x axis motor acceleration 
+void TC4_Handler(void);
+
+
+// -------------------------------- Y AXIS SETUP -------------------------------- //
+#define ACCEL_Y                 100
+#define VEL_MIN_Y               100
+#define VEL_MAX_Y               1000
+#define STEP_PIN_Y              22 // need to know the register for fast operation pin 5 - PB26 
+#define DIR_PIN_Y               23 // need to know the register for fast operation pin 6 - PA14
```
noaht11: Comments don't match the pin numbers

firmware/lib/Axis/src/Axis.h
```diff
@@ -0,0 +1,63 @@
...
+void isr_encoder_x();
+
+// isr to handle limit switch
+void isr_limit_switch_x();
+
+// isr for operating x axis motor velocity
+void TC3_Handler(void);
+
+// isr for operating x axis motor acceleration 
+void TC4_Handler(void);
```
noaht11: Don't need to declare these in the .h file if no-one else is going to use them. I think all 4 methods above are only used internally in Axis.cpp

firmware/lib/Axis/src/Axis.cpp
```diff
@@ -0,0 +1,264 @@
...
+        print("Encoder ISR count: ", axis_y.encoder.current);
+    }
+    last_interrupt_time = interrupt_time;
+}
+
+// isr to handle limit switch
+void isr_limit_switch_y()
+{   
+    static unsigned long last_interrupt_time = 0;
+    unsigned long interrupt_time = millis();
```
noaht11: Use `uint64_t`

firmware/lib/Axis/src/Axis.cpp
```diff
@@ -0,0 +1,264 @@
...
+        print("Encoder ISR count: ", axis_x.encoder.current);
+    }
+    last_interrupt_time = interrupt_time;
+}
+
+// isr to handle limit switch
+void isr_limit_switch_x()
+{   
+    static unsigned long last_interrupt_time = 0;
+    unsigned long interrupt_time = millis();
```
noaht11: Use ~`uint64_t`~ `uint32_t`.
`millis()` actually returns a `uint32_t`

noaht11:
Can you delete all the print statements, gonna clog up the serial port

firmware/lib/Axis/src/Axis.h
```diff
@@ -0,0 +1,63 @@
...
+// -------------------------------- X AXIS SETUP -------------------------------- //
+#define ACCEL_X                 100
+#define VEL_MIN_X               100
+#define VEL_MAX_X               1000
+#define STEP_PIN_X              5 // need to know the register for fast operation pin 5 - PC25 
+#define DIR_PIN_X               6 // need to know the register for fast operation pin 6 - PC24
+#define ENCODER_PIN_A_X         7
+#define ENCODER_PIN_B_X         8
+#define LIMIT_SW_HOME_PIN_X     9
+#define LIMIT_SW_FAR_PIN_X      10
```
noaht11: It seems like most if not all of these defines are only used in the Axis.cpp file. If that's the case they don't need to be defined in the header and can be moved to the cpp file.
Same for the Y defines.
If you think we will want to use them somewhere else in the future you can leave them here, but at least for things like pin numbers I don't imagine they will get used anywhere else.

firmware/lib/Axis/src/Axis.h
```diff
@@ -0,0 +1,63 @@
+#ifndef AXIS_H
+#define AXIS_H
+
+#include "Debug.h"
+#include "Kinematics.h"
```
noaht11: Only include in the header file, what's needed by the header file. If it's needed by the cpp file you can include it directly in the cpp file. So I think both of these includes can be moved to Axis.cpp

firmware/lib/Axis/src/Axis.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+    pinMode(LIMIT_SW_FAR_PIN_X, INPUT_PULLUP);
+    LimitSwitch ls_far_from_home = {LIMIT_SW_FAR_PIN_X, digitalRead(LIMIT_SW_FAR_PIN_X)};
+    
+    pinMode(DIR_PIN_X, OUTPUT);
+    pinMode(STEP_PIN_X, OUTPUT);
+    Axis axis_x_temp = {DIR_PIN_X, STEP_PIN_X, ACCEL_X, VEL_MAX_X, VEL_MIN_X, 
+                        {0, 0, 0}, 0, encoder_x, ls_home, ls_far_from_home,
+                        TC1, 0, 1, TC3_IRQn, TC4_IRQn};
+    return axis_x_temp;
```
noaht11: This method is basically identical to the y setup one. Can we do something like declare a:
```c
typedef struct {
    uint32_t encoderPinA, encoderPinB;
    uint32_t limitSwHomePin;
    // etc.
} AxisConfig;
```
Then have one method that you pass a `AxisConfig *` to

firmware/lib/Axis/src/Axis.h
```diff
@@ -0,0 +1,63 @@
...
+void TC0_Handler(void);
+
+// isr for operating x axis motor acceleration 
+void TC1_Handler(void);
+
+void setup_encoder_interrupts();
+
+void setup_ls_interrupts();
+
+#endif
```
noaht11: Minor thing but standard practice is to put a comment after an `#endif` for what it's ending (especially when it's really far away from the `#if`):
```c
#endif // AXIS_H
```

firmware/lib/Kinematics/src/Kinematics.h
```diff
@@ -0,0 +1,45 @@
...
+    LimitSwitch ls_far_from_home;
+    // necessary timer info
+    Tc *timer;
+    uint32_t channel_velocity;
+    uint32_t channel_accel;
+    IRQn_Type isr_velocity;
+    IRQn_Type isr_accel;
+    int dir;
+    bool homing;
+} Axis;
```
noaht11: Any reason this is defined in `Kinematics.h` rather than `Axis.h`?

firmware/lib/Kinematics/src/Kinematics.h
```diff
@@ -0,0 +1,45 @@
...
+#define VELOCITY_HOMING 200
+
+enum status{pressed, depressed};
+enum direction{positive, negative};
+enum segment{accelerate, hold, decelerate};
+
+typedef struct Axis
+{
+    int dir_pin;
+    int step_pin;
```
noaht11: Any reason these are `int`s rather than `uint32_t`s. `pinMode` and `digitalWrite` both take `uint32_t` as an argument

firmware/lib/Kinematics/src/Kinematics.h
```diff
@@ -0,0 +1,45 @@
...
+#include "Arduino.h"
+#include "Timer.h"
+#include "Encoder.h"
+#include "LimitSwitch.h"
+
+#define VELOCITY_HOMING 200
+
+enum status{pressed, depressed};
+enum direction{positive, negative};
+enum segment{accelerate, hold, decelerate};
```
noaht11: Should typedef all these enums:
```c
typedef enum { PRESSED, DEPRESSED } Status;
// etc.
```
Also think it would be worthwhile using UPPER_CASE_UNDERSCORES style so they don't look like local variables.
And use PascalCase for the name of the enum.

After typedefing them, you can use the name of the enum everywhere instead of int, so it's clear what you're actually expecting.

firmware/lib/Kinematics/src/Kinematics.h
```diff
@@ -0,0 +1,45 @@
...
+
+typedef struct Axis
+{
+    int dir_pin;
+    int step_pin;
+    uint32_t accel;
+    uint32_t vel_max;
+    uint32_t vel;
+    int32_t vel_profile_cur_trap[3]; // defined in counts accelerate, hold, decelerate
+    int tragectory_segment;
```
noaht11: After typedefing the enums, replace `int` with `Segment`. Same for the other ones.

firmware/lib/Axis/src/Axis.cpp
```diff
@@ -0,0 +1,264 @@
...
+    return axis_y_temp;
+}
+
+Axis axis_y = setup_axis_y();
+
+// isr to handle encoder of x axis
+void isr_encoder_y()
+{
+    static unsigned long last_interrupt_time = 0;
+    unsigned long interrupt_time = millis();
```
noaht11: Use ~`uint64_t`~ `uint32_t`.
`millis()` actually returns a `uint32_t`

firmware/lib/Axis/src/Axis.cpp
```diff
@@ -0,0 +1,264 @@
+#include "Axis.h"
+
+// -------------------------------- X AXIS SETUP -------------------------------- //
+Axis setup_axis_x()
```
noaht11: Given that all the pinMode configurations happen here, why not attach the interrupts as well?

firmware/lib/Axis/src/Axis.cpp
```diff
@@ -0,0 +1,264 @@
...
+    }
+    last_interrupt_time = interrupt_time;
+}
+
+// isr for operating x axis motor velocity
+void TC3_Handler(void)
+{   
+    // print("Desired count: ", axis_x.encoder.desired);
+    // print("Current count: ", axis_x.encoder.current);
+    TC_GetStatus(TC1, 0);
```
noaht11: Can you comment what TC_GetStatus does. I'm assuming it clears/acks the interrupt?

firmware/lib/Axis/src/Axis.cpp
```diff
@@ -0,0 +1,264 @@
...
+    // print("Current count: ", axis_x.encoder.current);
+    TC_GetStatus(TC1, 1);
+    // first accelerate
+    if (axis_x.tragectory_segment == accelerate) {
+        if (((axis_x.encoder.current < axis_x.encoder.desired) != axis_x.dir) && axis_x.vel < axis_x.vel_max) {
+            axis_x.vel++;
+            return;
+        }
+        else {
+            axis_x.tragectory_segment++;
```
noaht11: Are you incrementing an enum...I would classify that as relatively dangerous.
Especially since you didn't define numerical values for the enum constants.
I would say it's better to just hardcode what the next value is

firmware/lib/Axis/src/Axis.cpp
alexproskurin: Cool, will look into making it more concise

firmware/lib/Kinematics/src/Kinematics.h
alexproskurin: so, this isnt actually an enum, since i didnt think i could ++ enums i just kept it as int, but i compare it to enum for readability

firmware/lib/Axis/src/Axis.cpp
alexproskurin: Yeah, i kinda move things into modules super quickly and need to give more thought to the location of different things

firmware/lib/Axis/src/Axis.cpp
alexproskurin: Not incrementing enum here, that's why segment is defined as int

firmware/lib/Axis/src/Axis.cpp
```diff
@@ -0,0 +1,317 @@
...
+AxisConfig axis_x_config = {DIR_PIN_X, STEP_PIN_X, ENCODER_PIN_A_X, ENCODER_PIN_B_X, LIMIT_SW_HOME_PIN_X, LIMIT_SW_FAR_PIN_X,
+                            ACCEL_X, VEL_MAX_X, VEL_MIN_X, TC1, 0, 1, TC3_IRQn, TC4_IRQn, isr_encoder_x, isr_limit_switch_x};
+
+Axis setup_axis(AxisConfig *axis_config)
+{  
+    // Set up the timer interrupt
+    // TC1 channel 0, the IRQ for that channel and the desired frequency
+    pinMode(axis_config->encoder_pin_a, INPUT);
+    pinMode(axis_config->encoder_pin_b, INPUT);
+    Encoder encoder_x = {axis_config->encoder_pin_a, 10, 0};
```
noaht11: This no longer specific to x right? so just `encoder`?

firmware/lib/Axis/src/Axis.cpp
```diff
@@ -0,0 +1,317 @@
...
+#define DIR_PIN_X               6 // need to know the register for fast operation pin 6 - PC24
+#define ENCODER_PIN_A_X         7
+#define ENCODER_PIN_B_X         8
+#define LIMIT_SW_HOME_PIN_X     9
+#define LIMIT_SW_FAR_PIN_X      10
+
+AxisConfig axis_x_config = {DIR_PIN_X, STEP_PIN_X, ENCODER_PIN_A_X, ENCODER_PIN_B_X, LIMIT_SW_HOME_PIN_X, LIMIT_SW_FAR_PIN_X,
+                            ACCEL_X, VEL_MAX_X, VEL_MIN_X, TC1, 0, 1, TC3_IRQn, TC4_IRQn, isr_encoder_x, isr_limit_switch_x};
+
+Axis setup_axis(AxisConfig *axis_config)
```
noaht11: I guess I missed this on the first round, but generally speaking you don't want to return large structs from functions because they're being passed around by value. The standard practice would be to do something like this:
```c
void setup_axis(AxisConfig *axis_config, Axis *axis)
{
    axis->blah = blah;
    axis->blah2 = blah2;
    ...
}

AxisConfig axis_x_config = ...;
Axis axis_x;

void some_init_method()
{
    setup_axis(&axis_x_config, &axis_x);
}
```

firmware/lib/Axis/src/Axis.cpp
```diff
@@ -0,0 +1,317 @@
...
+                        axis_config->timer, axis_config->channel_velocity, axis_config->channel_velocity, axis_config->isr_velocity, axis_config->isr_accel};
+
+    // attach interrupts
+    attachInterrupt(digitalPinToInterrupt(axis_config->encoder_pin_a), axis_config->isr_encoder, RISING);
+    attachInterrupt(digitalPinToInterrupt(axis_config->ls_home), axis_config->isr_limit_switch, CHANGE);
+    attachInterrupt(digitalPinToInterrupt(axis_config->ls_far), axis_config->isr_limit_switch, CHANGE);
+    return axis_x_temp;
+}
+
+Axis axis_x = setup_axis(&axis_x_config);
```
noaht11: Don't love the idea of calling functions just in the global scope outside of any other function.
Especially ones that have side effects like setting pinModes and attaching interrupts.

Can we move these to some init function that can be called from main or just call them from main?

firmware/lib/Axis/src/Axis.cpp
```diff
@@ -0,0 +1,317 @@
...
+#define ENCODER_PIN_A_Y         24
+#define ENCODER_PIN_B_Y         25
+#define LIMIT_SW_HOME_PIN_Y     26
+#define LIMIT_SW_FAR_PIN_Y      27
+
+AxisConfig axis_y_config = {DIR_PIN_Y, STEP_PIN_Y, ENCODER_PIN_A_Y, ENCODER_PIN_B_X, LIMIT_SW_HOME_PIN_Y, LIMIT_SW_FAR_PIN_Y,
+                            ACCEL_Y, VEL_MAX_Y, VEL_MIN_Y, TC0, 0, 1, TC0_IRQn, TC1_IRQn, isr_encoder_y, isr_limit_switch_y};
+
+
+Axis axis_y = setup_axis(&axis_y_config);
```
noaht11: See comments about axis_x (applies here as well)

PR #7, opened 2020-03-09
Serial Communication Implementation

noaht11:
Implemented full framework for serial communication between the Arduino and the PC:

- `shared/TestStandComm`
  - Contains all the code for serial communication that is shared between the Arduino and the PC
  - Essentially all the protocol handling code
  - Does NOT include the serial device drivers or the application level code (sending specific messages)
- `shared_linux/LinuxSerialDevice`
  - Contains a linux serial device driver that allows the shared code to interface with a serial device on linux
- `shared_linux/TestStandCommHost`
  - Contains the application level code for sending specific messages from the host (PC) to the controller (Arduino)
- `firmware/lib/ArduinoSerialDevice`
  - Contains an Arduino serial device driver that allows the shared code to interface with a serial device on an Arduino
- `firmware/lib/TestStandCommController`
  - Contains the application level code for sending specific messages from the controller (Arduino) to the host (PC)
- `MessageTerminal`
  - Linux application to test sending/receiving messages to/from the Arduino
  - Provides a terminal style interface where you can run "commands" which send messages to the Arduino and process the responses
  - Currently only `ping` and `move` are actually implemented, with `move` just sending back the parameters and not actually calling the motor code
  - Steps to use:
    1. Navigate to the `MessageTerminal` folder
    2. Run `make`
    3. Run `./build/MessageTerminal /dev/ttyS#` where # is the serial port number the Arduino is connected to
    4. Enter commands such as: `ping` or `move 123 456 789 x pos`
    5. Type `exit` and hit enter to quit the program


MessageTerminal/MessageTerminal.cxx
```diff
@@ -0,0 +1,143 @@
...
+
+#include <iostream>
+#include <sstream>
+
+#include "LinuxSerialDevice.h"
+#include "TestStandCommHost.h"
+
+#define BAUD_RATE 115200
+
+#define BASIC_CMD(_name)                    \
```
chuandudx, line 13: Just looking to understand this structure. It is a macro where `_name` is replaced by the function `void _name(...)`?

MessageTerminal/MessageTerminal.cxx
```diff
@@ -0,0 +1,143 @@
...
+        iss >> word;
+        if (word == "pos") dir = DIR_POSITIVE;
+        else if (word == "neg") dir = DIR_NEGATIVE;
+        else break;
+
+        if (comm.move(accel, hold_vel, dist, axis, dir)) cout << "OK" << endl;
+        else cout << "ERROR" << endl;
+
+        if (comm.recv_message(5000) && comm.received_message().id == MSG_ID_LOG) {
+            printf("%s\n", &(comm.received_message().data[1]));
```
chuandudx, line 67: Is there a reason for mixing `cout` or `printf`? I also realize this is a test file so consistency/comments not a priority.

firmware/lib/Gantry/LimitSwitch.h
```diff
@@ -3,10 +3,15 @@
 
 #include "Arduino.h"
 
+typedef enum {
+    PRESSED = LOW,
+    RELEASED = HIGH
```
chuandudx, line 8: This is just preference, but we can also skip value assignment and use default enum values:
`typedef enum{PRESSED,RELEASED};`

firmware/lib/TestStandCommController/src/TestStandCommController.cxx
```diff
@@ -0,0 +1,36 @@
...
+#include "TestStandCommController.h"
+
+static uint8_t send_buf[MSG_DATA_LENGTH_MAX];
+
+TestStandCommController::TestStandCommController(SerialDevice& device) : TestStandComm(device)
+{
+    // Nothing else to do
+}
+
+bool TestStandCommController::log(LogLevel log_level, const char *fmt, ...)
```
chuandudx, line 11: What does `*fmt` stand for? What is `...`

shared/TestStandComm/Messages.h
```diff
@@ -0,0 +1,70 @@
...
+    LL_CRITICAL = 4
+} LogLevel;
+
+typedef enum {
+    STATUS_IDLE = 0,
+    STATUS_HOMING = 1,
+    STATUS_MOVING = 2,
+    STATUS_MEASURING = 3,
+    STATUS_LIMIT_REACHED = 4
+} Status;
```
chuandudx: Similar comment as before, we could use default enum values without explicitly assigning them since they increment from 0 onwards.

shared/TestStandComm/SerialDevice.h
```diff
@@ -0,0 +1,18 @@
...
+class SerialDevice
+{
+    public:
+        virtual bool ser_connect(uint32_t baud_rate) = 0;
+        virtual uint32_t ser_available() = 0;
+        virtual uint8_t ser_read() = 0;
+        virtual bool ser_write(uint8_t *data, uint32_t length) = 0;
+        virtual void ser_disconnect() = 0;
+
+        virtual uint64_t platform_millis() = 0;
```
chuandudx, line 15: Is `virtual` what allows `ArduinoSerialDevice` and `LinuxSerialDevice` to inherit from this class?

shared/TestStandComm/SerialTransport.h
```diff
@@ -0,0 +1,41 @@
...
+    MSG_SEG_END      = 5
+} MessageSegment;
+
+typedef struct {
+    MessageSegment current_segment;
+    uint8_t bytes_read;
+    uint8_t msg_length;
+    uint16_t crc;
+} PendingMessage;
+
```
chuandudx, line 22: Would it make sense to migrate `MessageSegment` and `PendingMessage` also to `Messages.h`?

shared_linux/LinuxSerialDevice/LinuxSerialDevice.cxx
```diff
@@ -0,0 +1,112 @@
...
+#include <sys/ioctl.h> // ioctl()
+#include <sys/time.h>
+
+LinuxSerialDevice::LinuxSerialDevice()
+{
+    // Nothing to do
+}
+
+void LinuxSerialDevice::set_device_file(const char *device_file)
+{
```
chuandudx, line 21: Can setting the device_file be handled by the constructor?

shared_linux/LinuxSerialDevice/LinuxSerialDevice.cxx
```diff
@@ -0,0 +1,112 @@
...
+void LinuxSerialDevice::ser_disconnect()
+{
+    close(this->serial_port);
+}
+
+uint64_t LinuxSerialDevice::platform_millis()
+{
+    struct timeval tval;
+    gettimeofday(&tval, NULL);
+    return (tval.tv_usec / 1000);
```
chuandudx, line 111: should we use `round` otherwise the time is truncated.

shared_linux/TestStandCommHost/TestStandCommHost.cxx
```diff
@@ -0,0 +1,60 @@
+#include <stdio.h>
+#include "TestStandCommHost.h"
+
+#define GET_BYTE(_x, _n) (uint8_t)(((_x) >> ((_n)*8)) & 0x00FF)
```
chuandudx: migrate to `macros.h`?

shared_linux/TestStandCommHost/TestStandCommHost.cxx
```diff
@@ -0,0 +1,60 @@
...
+{
+    // Transmit in big endian order
+    uint8_t data[8];
+    data[0] = GET_BYTE(accel, 1);
+    data[1] = GET_BYTE(accel, 0);
+    data[2] = GET_BYTE(hold_vel, 1);
+    data[3] = GET_BYTE(hold_vel, 0);
+    data[4] = GET_BYTE(dist, 1);
+    data[5] = GET_BYTE(dist, 0);
+    data[6] = (uint8_t)axis;
```
chuandudx: Replace `1` with `AXIS_Y` and `0` with `AXIS_X` for clarity.

chuandudx (approved):
Overall the structure is quite clear. I had several comments mainly to understand the reason for some implementations for my own knowledge. More comments for the functions throughout, in general, would be helpful for documentation and readability.

firmware/lib/Gantry/LimitSwitch.h
noaht11: By definition, these have to be equal to LOW and HIGH, so I'd rather explicitly declare that rather than rely on the implicit dependency that PRESSED == 0 and LOW == 0 therefore PRESSED == LOW

firmware/lib/TestStandCommController/src/TestStandCommController.cxx
noaht11: fmt is short for format, I think it's a pretty standard abbreviation I've seen in C code

... is varargs (variable arguments). It basically lets you pass any number of arguments you want to the function. This is how printf works (where you pass a format, followed by as many variables as you want).

shared/TestStandComm/Messages.h
noaht11: Yeah this one I guess doesn't need them defined

shared/TestStandComm/SerialDevice.h
noaht11: Yep.
And ` = 0` is what forces them to implement the function

shared_linux/TestStandCommHost/TestStandCommHost.cxx
noaht11: Yeah, that's in progress in what I'm working on right now, gonna push a new commit with it

shared_linux/TestStandCommHost/TestStandCommHost.cxx
noaht11: 1 and 0 there are referring to first byte and second byte of the distance, not to the axis

MessageTerminal/MessageTerminal.cxx
noaht11: I spent about 10 seconds trying to get `cout` to print formatted stuff and it didn't work so I just switched to `printf` whenever I needed to format the output.
I think maybe I'll just use `printf` everywhere.

MessageTerminal/MessageTerminal.cxx
noaht11: It gets replaced with the entire function, so:

`BASIC_CMD(ping);`

gets replaced by exactly this:

```c
void ping(istringstream& iss)
{
    if (comm.ping()) cout << "OK" << endl;
    else cout << "ERROR" << endl;
};
```

shared_linux/LinuxSerialDevice/LinuxSerialDevice.cxx
noaht11: I didn't bother cause I don't particularly care if the timeout is off by one millisecond. I don't think it needs to be precise (it's just a waiting time).

shared/TestStandComm/Messages.h
noaht11: Moved all of these enums into shared_defs.h and removed the explicit numbers assigned to them

shared_linux/TestStandCommHost/TestStandCommHost.cxx
noaht11: Moved this into macros.h. Also changed the macros to be
`NTOHS` = Network-to-host short (16-bit)
`NTOHL` = Network-to-host long (32-bit)
 `HTONS` = Host-to-network short (16-bit)
 `HTONL` = Host-to-network long (32-bit)

shared_linux/LinuxSerialDevice/LinuxSerialDevice.cxx
noaht11: I didn't want it in the constructor because I wanted to declare a global variable for the LinuxSerialDevice and you have to provide the constructor arguments when you declare the global variable, but we don't get this argument until main is running.

shared/TestStandComm/SerialTransport.h
noaht11: I'd rather keep theme here, since they're exclusively used by SerialTransport.h and SerialTransport.cxx.
Someone who uses Messages.h doesn't need access to these structs

MessageTerminal/MessageTerminal.cxx
noaht11: Leaving as is for now, can come back to this later

PR #8, opened 2020-03-12
Midas control

chuandudx:
Implemented functions to send motor move command to the Arduino Due through the MIDAS frontend. MIDAS allows user to input the following parameters
- `Destination`, `Velocity`, `Acceleration` in both x and y direction

`motorCalculation`
 - Converts user input in units based in millimeters to motor encoder counts
- determines whether to move gantry in positive or negative direction

`feArduino`
-  obtains values specified by the user from the MIDAS interface
- error checks to ensure the destination, velocity, and acceleration values are valid
- send motor command to the Arduino Due

TODO: may need to remove user's ability to configuration acceleration.


noaht11:
Abandoning in favor of #11

PR #9, opened 2020-03-20
Thermistor library for temperature measurement

chuandudx:
ThermistorUSP10876 library converts analog readings from 10k thermistors to temperature in degrees Celsius.
- Arduino Due has 12-bit ADC resolution which is enabled by `analogReadResolution(12)` in `main.cpp`
- The resistance to temperature conversion uses Steinhart-hart equation constants estimated from the manufacturer provided lookup table.
- `main.cpp` contains serial monitor output of temperature reading in csv format for testing. This does not need to be merged.
- `readAveragedTemperature()` takes 5 consecutive analog readings and averages them before calculating the temperature. This was implemented in case we need to account for noise through averaging. However, under regular conditions, this function does not appear to have more consistency compared to `readTemperature()` which takes a single reading. 
- Further testing is required under noise conditions from motor. In the ambient air, all five tested thermistors output within 0.5 degrees of each other.


firmware/lib/thermistorProbe/thermistorUSP10976.h
```diff
@@ -0,0 +1,44 @@
+#pragma once
+#ifndef THERMISTOR_USP10976_H
+#define THERMISTOR_USP10976_H
+
+class Thermistor_10k
```
noaht11: Can we name the file the same as the class it contains. Also would prefer no underscores in class names, just CamelCase

noaht11: Can we name the file the same as the class it contains. Also would prefer no underscores in class names, just CamelCase

firmware/lib/thermistorProbe/thermistorUSP10976.cpp
```diff
@@ -0,0 +1,55 @@
...
+ * Thermistor 10k temperature measurements.
+ */
+#include <Arduino.h>
+#include "thermistorUSP10976.h"
+
+/**
+ * 
+ * @param pin - Arduino Due pin
+ */
+Thermistor_10k::Thermistor_10k(int pin, double seriesResistor){
```
noaht11: nit: Bracket on next line for functions pls :)

noaht11: Also `uint32_t` for the pin data type. (That's the type used in the arguments to `pinMode`, `digitalRead` etc.)

firmware/lib/thermistorProbe/thermistorUSP10976.h
```diff
@@ -0,0 +1,44 @@
...
+{
+    public:
+        // const double resolutionLim = 1023.0; // 10-bit Arduino Uno
+        const double resolutionLim = 4094.0; // 12-bit Arduino Due
+        // const float seriesResistor = 10000.0; 
+        const double vRef = 3.3;
+        const double c1 = 0.001127354682, c2 = 0.0002343978227, c3 = 0.00000008674847738; //Steinhart-hart equation constants
+        const double kelvin_to_celsius = 273.15;
+        static const int numSamples = 5;
```
noaht11: Do all these consts need to be public? Does anyone outside the class use them?

firmware/lib/thermistorProbe/thermistorUSP10976.h
```diff
@@ -0,0 +1,44 @@
...
+
+        int vLevel;
+        double seriesResistorVal;
+        double vTherm;
+        double logRTherm,RTherm, Tk, Tc;
```
noaht11: As far as I can tell, there's no reason for these to be member variables? Can they not just be local variables in the functions that use them?
(Referring to `vLevel`, `vTherm`, `logRTherm`, `RTherm`, `Tk`, `Tc` here)

firmware/lib/thermistorProbe/thermistorUSP10976.h
```diff
@@ -0,0 +1,44 @@
...
+        // const double resolutionLim = 1023.0; // 10-bit Arduino Uno
+        const double resolutionLim = 4094.0; // 12-bit Arduino Due
+        // const float seriesResistor = 10000.0; 
+        const double vRef = 3.3;
+        const double c1 = 0.001127354682, c2 = 0.0002343978227, c3 = 0.00000008674847738; //Steinhart-hart equation constants
+        const double kelvin_to_celsius = 273.15;
+        static const int numSamples = 5;
+
+        int vLevel;
+        double seriesResistorVal;
```
noaht11: Couldn't `seriesResistorVal` be private. Seems very much like internal information for the class

firmware/lib/thermistorProbe/thermistorUSP10976.h
```diff
@@ -0,0 +1,44 @@
...
+
+class Thermistor_10k
+{
+    public:
+        // const double resolutionLim = 1023.0; // 10-bit Arduino Uno
+        const double resolutionLim = 4094.0; // 12-bit Arduino Due
+        // const float seriesResistor = 10000.0; 
+        const double vRef = 3.3;
+        const double c1 = 0.001127354682, c2 = 0.0002343978227, c3 = 0.00000008674847738; //Steinhart-hart equation constants
+        const double kelvin_to_celsius = 273.15;
```
noaht11: Can these consts also be static?

firmware/lib/thermistorProbe/thermistorUSP10976.h
```diff
@@ -0,0 +1,44 @@
...
+        /**
+         * Obtains analog thermistor reading and converts it to temperature in Celsius
+         * @return temperature in Celsius
+         */
+        double readTemperature();
+        /**
+         * Obtains five analog thermistor reading averages them and converts it to temperature in Celsius
+         * @return temperature in Celsius
+         */
+        double readAveragedTemperature();
```
noaht11: Have you noticed a difference between `readTemperature` and `readAveragedTemperature` results. Just curious since you're only averaging across 50 ms, wasn't sure if it would make a huge difference

firmware/lib/thermistorProbe/thermistorUSP10976.h
```diff
@@ -0,0 +1,44 @@
...
+        const double c1 = 0.001127354682, c2 = 0.0002343978227, c3 = 0.00000008674847738; //Steinhart-hart equation constants
+        const double kelvin_to_celsius = 273.15;
+        static const int numSamples = 5;
+
+        int vLevel;
+        double seriesResistorVal;
+        double vTherm;
+        double logRTherm,RTherm, Tk, Tc;
+        
+        int vLevelSamples[numSamples];
```
noaht11: Does this need to be a member variable?

firmware/lib/thermistorProbe/thermistorUSP10976.h
```diff
@@ -0,0 +1,44 @@
+#pragma once
+#ifndef THERMISTOR_USP10976_H
+#define THERMISTOR_USP10976_H
```
noaht11: `#pragma once` is redundant with the header guard (`ifndef`, `define`) you have right after it (they each ensure the header file is only included once, so you only need one of them)

We've been going with header guards in all the other files, so lets stick to that for consistency.

tldr: remove `#pragma once`

noaht11 (changes requested):
I'm thinking for this PR let's merge only your thermistor library (i.e. revert the changes to main.cpp and platformio.ini). That way there's no conflicts in this PR and you can do a second PR based off the latest develop and modify main.cpp to instantiate your thermistor objects. What do you think of that approach?

firmware/lib/thermistorProbe/thermistorUSP10976.h
chuandudx: No, `readAveragedTemperature()` did not really have a difference over `readTemperature`. Also, since testing was done in a noise free condition, `readTemperature` already gave pretty consistent data. Since we can change the number of samples averaged so it can be used in the future if noise proves to be an issue in the gantry environment.

firmware/lib/thermistorProbe/thermistorUSP10976.h
chuandudx: should `resolutionLim` also be static? What is the advantage for a variable also being `static` in addition to being `const`.

chuandudx: `'constexpr' needed for in-class initialization of static data member 'const double Thermistor10k::c2' of non-integral type [-fpermissive]`
This is the error when I tried to make the above static.

firmware/lib/thermistorProbe/thermistorUSP10976.h
chuandudx: Followed your suggestion and moved member variables to either local or private when appropriate.

firmware/lib/thermistorProbe/thermistorUSP10976.h
chuandudx: Good suggestion!

firmware/lib/thermistorProbe/thermistor10k.h
```diff
@@ -0,0 +1,39 @@
+#ifndef THERMISTOR_USP10976_H
+#define THERMISTOR_USP10976_H
```
noaht11: Rename the header to guard to match the file/class name (`THERMISTOR_10K_H`)

firmware/lib/thermistorProbe/thermistor10k.cpp
```diff
@@ -0,0 +1,60 @@
...
+    int vLevelSamples[numSamples];
+
+    for(int i = 0; i < numSamples; i++){
+        vLevelSamples[i] = analogRead(thermistorPin);
+        delay(10);
+    }
+
+    for(int i = 0; i < numSamples; i++){
+        averageVLevel += vLevelSamples[i];
+    }
```
noaht11: Don't need the `vLevelSamples` array, and only need one loop here.
On each iteration just add to `averageVLevel` the latest reading, then at the end divide by `numSamples` like you already have.

firmware/lib/thermistorProbe/thermistor10k.cpp
```diff
@@ -0,0 +1,60 @@
...
+ * @return temperature in Celsius
+ */
+double Thermistor10k::readAveragedTemperature(int numSamples)
+{
+    double averageVLevel = 0.0;
+    int vLevelSamples[numSamples];
+
+    for(int i = 0; i < numSamples; i++){
+        vLevelSamples[i] = analogRead(thermistorPin);
+        delay(10);
```
noaht11: At some point we'll have to re-factor this to not use delays because any delays in the main loop might cause us to miss serial data. For now we'll leave it as is, so we can get this merged but just something to be aware of.

firmware/lib/thermistorProbe/thermistor10k.h
```diff
@@ -0,0 +1,39 @@
...
+        double readAveragedTemperature(int numSamples);
+        
+    private:
+        uint32_t thermistorPin; 
+        double seriesResistorVal;       
+        // const double resolutionLim = 1023.0; // 10-bit Arduino Uno
+        const double resolutionLim = 4094.0; // 12-bit Arduino Due
+        const double vRef = 3.3;
+        const double c1 = 0.001127354682, c2 = 0.0002343978227, c3 = 0.00000008674847738; //Steinhart-hart equation constants
+        const double kelvin_to_celsius = 273.15;
```
noaht11: Okay I just went down a giant rabbit hole of c++ weirdness. Essentially, you can make these static by changing `const` to `constexpr`.
The reason to make them static is because without static, each instance of this class will have a copy of these variables even though they are the same for all instances by definition/design.
The reason you need `constexpr` instead of `const` is a weird c++ thing that has to do with the fact that they are `double`s and not `int`s and `double`s get initialized differently.

firmware/lib/thermistorProbe/thermistor10k.cpp
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Obtains analog thermistor reading and converts it to temperature in Celsius
+ * @return temperature in Celsius
+ */
+double Thermistor10k::readAveragedTemperature(int numSamples)
+{
+    double averageVLevel = 0.0;
+    int vLevelSamples[numSamples];
+
+    for(int i = 0; i < numSamples; i++){
```
noaht11: nit: space in between `)` and `{`

firmware/lib/thermistorProbe/thermistor10k.h
```diff
@@ -0,0 +1,39 @@
...
+         * Obtains five analog thermistor reading averages them and converts it to temperature in Celsius
+         * @return temperature in Celsius
+         */
+        double readAveragedTemperature(int numSamples);
+        
+    private:
+        uint32_t thermistorPin; 
+        double seriesResistorVal;       
+        // const double resolutionLim = 1023.0; // 10-bit Arduino Uno
+        const double resolutionLim = 4094.0; // 12-bit Arduino Due
```
noaht11: Shouldn't this be `4095`? (2^12 - 1 = 4095)

noaht11 (changes requested):
A couple minor issues left. Also don't forget to revert `main.cpp` and `platformio.ini` before merging. Maybe tuck away a copy of main.cpp for reference for the second PR where you integrate this.

To revert the files you can do:
```
git checkout 130d71bc279643f15ea779540546e6a40260ffe8 .\src\main.cpp .\platformio.ini
```
before you commit.

Last thing, this is purely for consistency's sake, but can we do capital first letters on the folder and file names:

- `thermistorProbe` -> `ThermistorProbe`
- `thermistor10k.h` -> `Thermistor10k.h`
- `thermistor10k.cpp` -> `Thermistor10k.cpp`

firmware/lib/thermistorProbe/thermistorUSP10976.h
noaht11: See my comment in the second review below

firmware/lib/thermistorProbe/thermistor10k.cpp
chuandudx: Good point!

firmware/lib/thermistorProbe/thermistor10k.h
chuandudx: Yes, good catch!

firmware/lib/thermistorProbe/thermistor10k.h
```diff
@@ -0,0 +1,40 @@
...
+        double readAveragedTemperature(int numSamples);
+        
+    private:
+        uint32_t thermistorPin; 
+        double seriesResistorVal;       
+        static constexpr double resolutionLim = 4095.0; // 12-bit Arduino Due
+        static constexpr double vRef = 3.3;
+        static constexpr double c1 = 0.001127354682, c2 = 0.0002343978227, c3 = 0.00000008674847738; //Steinhart-hart equation constants
+        static constexpr double kelvin_to_celsius = 273.15;
+        double convertToTemp(int vLevel);
```
noaht11: Super minor nit picky thing, but can we re-order these to:

```c++
private:
        static constexpr double resolutionLim = 4095.0; // 12-bit Arduino Due
        static constexpr double vRef = 3.3;
        static constexpr double c1 = 0.001127354682, c2 = 0.0002343978227, c3 = 0.00000008674847738; //Steinhart-hart equation constants
        static constexpr double kelvin_to_celsius = 273.15;

        uint32_t thermistorPin;
        double seriesResistorVal;

        double convertToTemp(int vLevel);
```

noaht11 (approved):
One last super tiny comment.
Also did you do the capital first letter in the filenames (doesn't look like it got changed).

Otherwise good to go!

firmware/lib/thermistorProbe/thermistor10k.h
chuandudx: Oh weird, I definitely made the change, but did not come through properly. I will try again.

PR #10, opened 2020-03-22
Thermistor initialization in main.cpp

chuandudx:
* Initialized 5 thermistor objects in `main.cpp` for temperature readings for ambient, motor1, mPMT, motor2, and optical box. 
* Modified `ThermistorProbe` constructor to have `seriesResistorVal` to be `static constexpr double`. Previously, we needed to measure each resistor to obtain precise calibration. Now with precision resistors (10k ± 0.1%) variation in resistance is no longer a significant contributing factor to temperature accuracy.

### DIFF
--- a/firmware/lib/ThermistorProbe/Thermistor10k.cpp
+++ b/firmware/lib/ThermistorProbe/Thermistor10k.cpp
@@ -8,10 +8,9 @@
 /**
  * @param pin - Arduino Due pin
  */
-Thermistor10k::Thermistor10k(uint32_t pin, double seriesResistor)
+Thermistor10k::Thermistor10k(uint32_t pin)
 {
     this->thermistorPin = pin;
-    this->seriesResistorVal = seriesResistor; // measure actual resistance of the 10 kOhm series resistor
 }
 /**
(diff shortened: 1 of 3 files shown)


noaht11 (approved):
Do you need to call `pinMode` for each of the thermistor pins? Even if it defaults to the right mode maybe it's worth calling anyway to be explicit? Not sure what the convention is.

Other than that, looks good

chuandudx:
> Do you need to call `pinMode` for each of the thermistor pins? Even if it defaults to the right mode maybe it's worth calling anyway to be explicit? Not sure what the convention is.
> 
> Other than that, looks good

I did not explicitly call `pinMode` based on this source 

> For example, when calling analogRead(), an analog input pin is automatically changed from a digital input (or output) into an analog input. For this reason, it isn’t necessary to call the pinMode function on the pin.
https://www.baldengineer.com/tips/arduino-pinmode-on-analog-inputs

The same source also mentions one could prefer to declare `pinMode(uint32_t pin, INPUT)` anyway for clarity so it agrees with your suggestion and I will make the change.

PR #11, opened 2020-03-22
Midas Control v2

noaht11:
Implemented functions to send motor move command to the Arduino Due through the MIDAS frontend. MIDAS allows user to input the following parameters
- `Destination`, `Velocity`, `Acceleration` in both x and y direction

`motorCalculation`
 - Converts user input in units based in millimeters to motor encoder counts
- determines whether to move gantry in positive or negative direction

`feArduino`
-  obtains values specified by the user from the MIDAS interface
- error checks to ensure the destination, velocity, and acceleration values are valid
- send motor command to the Arduino Due

TODO: may need to remove user's ability to configuration acceleration.

### DIFF
--- a/MidasFrontend/feArduino.cxx
+++ b/MidasFrontend/feArduino.cxx
@@ -141,11 +141,24 @@ INT start_move(){
   // Get the destination position
   std::string destpath = path + "/Destination";
   float destination[2] = {0,0};
-  int size = sizeof(destination);
-  int status = db_get_value(hDB, 0, destpath.c_str(), &destination, &size, TID_FLOAT, TRUE);
-        
-  printf("Moving to position X=%f, Y=%f\n",destination[0],destination[1]);
+  int size_dest = sizeof(destination);
+  int status_dest = db_get_value(hDB, 0, destpath.c_str(), &destination, &size_dest, TID_FLOAT, TRUE);
 
+  // Get the velocity
+  std::string velpath = path + "/Velocity";
+  float velocity[2] = {0,0};
+  int size_vel = sizeof(velocity);
+  int status_vel = db_get_value(hDB, 0, velpath.c_str(), &velocity, &size_vel, TID_FLOAT, TRUE);
+
+  // Get the acceleration
+  std::string accelpath = path + "/Acceleration";
+  float acceleration[2] = {0,0};
+  int size_accel = sizeof(acceleration);
+  int status_accel = db_get_value(hDB, 0, accelpath.c_str(), &acceleration, &size_accel, TID_FLOAT, TRUE);
+
+  printf("Moving to position P_x=%f, P_y=%f\n",destination[0],destination[1]);
+  printf("Moving with velocity V_x=%f, V_y=%f\n",velocity[0],velocity[1]);
+  printf("Moving with acceleration A_x=%f, A_y=%f\n",acceleration[0],acceleration[1]);
   // TOFIX: instruct the Arduino to move to the specified destination at specified speed.
   
   for(int i = 0; i < 5; i++){
--- a/MidasFrontend/feArduino.cxx
+++ b/MidasFrontend/feArduino.cxx
@@ -26,9 +26,9 @@ Arduino motor control for mPMT test stand.
 #define  EQ_NAME   "ARDUINO"
 #define  EQ_EVID   1
 #define  EQ_TRGMSK 0x1111
-#define  TIME_OUT  5000
+#define  MSG_RECEIVE_TIMEOUT  5000
 
-#define BAUD_RATE 115200
+#define  BAUD_RATE 115200
 
 // error checking user inputs
 const float gantry_x_min_mm = 0.0;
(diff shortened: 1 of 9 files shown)


MidasFrontend/feArduino.cxx
```diff
@@ -7,18 +7,42 @@ Arduino motor control for mPMT test stand.
...
-#include "unistd.h"
 #include "time.h"
 #include "sys/time.h"
-#include <stdint.h>
-
 
+#include "LinuxSerialDevice.h"
+#include "TestStandCommHost.h"
+#include "motorCalculation.h"
+#include "macros.h"
```
noaht11, line 24: Nicely organized includes :)

MidasFrontend/feArduino.cxx
```diff
@@ -7,18 +7,42 @@ Arduino motor control for mPMT test stand.
...
 
 #define  EQ_NAME   "ARDUINO"
 #define  EQ_EVID   1
 #define  EQ_TRGMSK 0x1111
+#define  TIME_OUT  5000
```
noaht11: Name could be more descriptive, something like `MSG_RECEIVE_TIMEOUT` or similar

MidasFrontend/feArduino.cxx
```diff
@@ -7,18 +7,42 @@ Arduino motor control for mPMT test stand.
...
+
+// error checking user inputs
+const float gantry_x_min_mm = 0.0;
+const float gantry_x_max_mm = 1200.0; // max rail is 1219 mm
+const float gantry_y_min_mm = 0.0;
+const float gantry_y_max_mm = 1200.0;
+const float vel_min_mm_s = 0.0; 
+const float vel_max_mm_s = 10.0;
+const float accel_min_mm_s_2 = 0.0; // acceleration cannot be zero
+const float accel_max_mm_s_2 = 7.0;
```
noaht11: I realize I forgot to mention this in the other PR, but I think it would be good to make consts all uppercase with underscores. What do you think?

MidasFrontend/feArduino.cxx
```diff
@@ -126,8 +146,10 @@ BOOL gStartMove;
 HNDLE handleHome;
 HNDLE handleMove;
 
-
 INT start_move(){
+    uint16_t accel, hold_vel, dist;
```
noaht11: You can go ahead and update these to uint32_t now

MidasFrontend/feArduino.cxx
```diff
@@ -138,15 +160,127 @@ INT start_move(){
   path += EQ_NAME;
   path += "/Settings";
 
-  // Get the destination position
+  // Get the destination position (absolute distance)
   std::string destpath = path + "/Destination";
```
noaht11: You have a `using namespace std` declaration at the top, so you shouldn't need the `std::` here. Same for all the ones below

MidasFrontend/motorCalculation.cxx
```diff
@@ -0,0 +1,51 @@
+#include "motorCalculation.h"
+
+using namespace std;
+
+//retreive values from memory
+float curr_pos_cts;
+//user inputs
+float dest_mm, vel_mm_s, accel_mm_s2;
```
noaht11: Pretty sure these variables can be deleted, since they're passed in as arguments to the functions

MidasFrontend/motorCalculation.h
```diff
@@ -0,0 +1,18 @@
...
+#define MOTOR_CALCULATION_H
+
+#include <cmath>
+#include <iostream>
+#include "Messages.h" 
+
+const float encoder_cpr = 300.0;
+const float lead_screw_pitch_mm = 8.0;
+const float gear_ratio = 26.0 + 103.0/121.0;
+const float mm_cts_ratio =  lead_screw_pitch_mm/(encoder_cpr*gear_ratio);
```
noaht11: If these are only used by `motorCalculation.cxx` then you can move them into that file. No need to share them with everyone who includes this header file.
Also all caps?

noaht11 (commented):
A few minor things to fix. I ended up pushing up two new commits while I was testing getting MIDAS running so you might notice a few changes (nothing significant).
You should be able to make these fixes and build within WSL, you just won't be able to run. I'm working on getting a VM setup

MidasFrontend/feArduino.cxx
```diff
@@ -177,18 +308,27 @@ INT start_home(){
...
+    puts(argv[i]);
+  }
+
+  if (argc != 2) {
+    printf("\nusage: %s <serial device file>\n\nexample:\n    %s /dev/ttyACM0\n\n", argv[0], argv[0]);
+    return 0;
+  }
+
+  device.set_device_file(argv[1]);
+  if (!device.ser_connect(BAUD_RATE)) return 1;
```
noaht11: Should probably return `FE_ERR_HW` here instead

MidasFrontend/motorCalculation.h
chuandudx: Migrated! Just wondering though, does the header file not serve to have constants so that they are easy to find, and the `.cpp` may be cleaner?

MidasFrontend/motorCalculation.h
```diff
@@ -0,0 +1,15 @@
+#ifndef MOTOR_CALCULATION_H
+#define MOTOR_CALCULATION_H
+
+#include <cmath>
+#include <iostream>
+
+#include "Messages.h"
```
noaht11: I think this include isn't needed anymore

noaht11 (commented):
One last tiny comment then good to go, sorry missed it on my first pass

PR #13, opened 2020-03-26
Large code restructuring for cleanup, top-level architecture and serial reliability

noaht11:
- Created mPMTTestStand class and moved most of the main.cpp code into there
  - mPMTTestStand accepts a list of IO configurations (pin numbers, serial devices etc.) to setup all the other objects
  - The motor control pins are included but not currently used, will come in a future PR

- Modified the startup procedure, so the Arduino pings the host until the host responds
  - The host equivalently waits for a ping from the Arduino before sending any other commands
  - PING is now a two-way message to allow for this
  - Implemented a flush feature in the serial device libraries so that the extra ping messages can be flushed out

- Misc. serial reliability improvements
  - pending message resets even if the END byte is incorrect
  - session will not ACK an ACK and ignore it instead

- Enabled DEBUG

### DIFF
--- a/MessageTerminal/MessageTerminal.cxx
+++ b/MessageTerminal/MessageTerminal.cxx
@@ -10,11 +10,16 @@
 
 #define BAUD_RATE 115200
 
-#define BASIC_CMD(_name)                    \
-void _name(istringstream& iss)              \
-{                                           \
-    if (comm._name()) cout << "OK" << endl; \
-    else cout << "ERROR" << endl;           \
+#define BASIC_CMD(_name)                     \
+void _name(istringstream& iss)               \
+{                                            \
+    if (comm._name()) {                      \
+        cout << "OK" << endl;                \
+    }                                        \
+    else {                                   \
+        printf("ERR: Received MSG ID: %d\n", \
+                comm.received_message().id); \
+    }                                        \
 }
 
 using namespace std;
@@ -60,8 +65,12 @@ void move(istringstream& iss)
         else if (word == "neg") dir = DIR_NEGATIVE;
         else break;
 
-        if (comm.move(accel, hold_vel, dist, axis, dir)) cout << "OK" << endl;
-        else cout << "ERROR" << endl;
+        if (comm.move(accel, hold_vel, dist, axis, dir)) {
+            cout << "OK" << endl;
+        }
+        else {
+            printf("ERR: Received MSG ID: %d\n", comm.received_message().id);
+        }
 
         if (comm.recv_message(5000) && comm.received_message().id == MSG_ID_LOG) {
             printf("%s\n", &(comm.received_message().data[1]));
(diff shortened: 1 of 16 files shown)


MessageTerminal/MessageTerminal.cxx
```diff
@@ -60,8 +65,12 @@ void move(istringstream& iss)
         else if (word == "neg") dir = DIR_NEGATIVE;
         else break;
 
-        if (comm.move(accel, hold_vel, dist, axis, dir)) cout << "OK" << endl;
-        else cout << "ERROR" << endl;
+        if (comm.move(accel, hold_vel, dist, axis, dir)) {
+            cout << "OK" << endl;
+        }
+        else {
+            printf("ERR: Received MSG ID: %d\n", comm.received_message().id);
```
chuandudx, line 72: Use either `cout` or `printf` for consistency throughout

firmware/lib/ArduinoSerialDevice/src/ArduinoSerialDevice.cxx
```diff
@@ -11,6 +11,17 @@ bool ArduinoSerialDevice::ser_connect(uint32_t baud_rate)
     return true;
 }
 
+void ArduinoSerialDevice::ser_flush()
+{
+    // Wait for 10 ms for last bits of data
+    delay(10);
```
chuandudx, line 17: I realize this is a very short delay, but do we foresee any issue this might cause?

noaht11: It's less than ideal, but this function is generally only called once at setup if at all, so it shouldn't cause any problems and I don't have an alternative.

MessageTerminal/MessageTerminal.cxx
noaht11: My "consistency" is currently `printf` for formatted output, `cout` for non-formatted. Is that consistent-ish-enough?

chuandudx: Didn't realize this was the reasoning. Yes, that makes sense now thank you

firmware/src/mPMTTestStand.h
```diff
@@ -0,0 +1,63 @@
...
+        void handle_home();
+        void handle_move();
+
+    public:
+        mPMTTestStand(const mPMTTestStandIO& io);
+        void setup();
+        void execute();
+};
+
+#endif // CMD_HANDLER_H
```
chuandudx, line 63: Really clear pin organization!

firmware/src/main.cpp
```diff
@@ -1,99 +1,58 @@
...
-                break;
-            case MSG_ID_MOVE:
-                handle_move();
-                break;
-        }
+    test_stand.execute();
+
+    // Blink an LED
+    if ((millis() - blink_start) >= 500) {
+        blink_state = !blink_state;
```
chuandudx, line 54: What is the meaning of `blink_state = !blink_state`?

noaht11: Toggle the `blink_state` (i.e. set it to the opposite of what it was before)

firmware/src/mPMTTestStand.h
noaht11: Thanks :) that was one of the objectives of these changes

PR #15, opened 2020-03-29
Pseudo gantry

alexproskurin:


PseudoGantry/src/main.cpp
```diff
@@ -0,0 +1,38 @@
...
+{
+  pseudo_encoder_x.channel_a_out ^= HIGH;
+}
+void isr_encoder_y()
+{
+  pseudo_encoder_y.channel_a_out ^= HIGH;
+}
+
+void setup() {
+  pseudo_encoder_x = {MOTOR_AXIS_Y, ENCODER_X_OUT};
```
noaht11: Typo, should be `MOTOR_AXIS_X` I'm guessing

PseudoGantry/lib/PseudoGantry/PseudoAxis.h
```diff
@@ -0,0 +1,14 @@
+#ifndef PSEUDO_AXIS_H
+#define PSEUDO_AXIS_H
+
+#include "Arduino.h"
+#include "PseudoEncoder.h"
+
+typedef struct PseudoAxis
+{
+    PseudoEncoder encoder;
+    uint32_t axis_length;  // specified in counts
```
noaht11: Maybe just name it `axis_length_counts` so we know everywhere we use it

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
```diff
@@ -0,0 +1,2 @@
+#include "PseudoAxis.h"
+
```
noaht11, line 2: Empty file?

PseudoGantry/src/main.cpp
```diff
@@ -0,0 +1,38 @@
...
+#define AXIS_LENGTH_Y 1000000 
+
+PseudoEncoder pseudo_encoder_x;
+PseudoEncoder pseudo_encoder_y;
+PseudoAxis pseudo_axis_x;
+PseudoAxis pseudo_axis_y;
+
+void isr_encoder_x()
+{
+  pseudo_encoder_x.channel_a_out ^= HIGH;
```
noaht11: Isn't this a pin number? Do you mean to be toggling the output on that pin?

PseudoGantry/lib/PseudoGantry/PseudoEncoder.cpp
```diff
@@ -0,0 +1,9 @@
+#include "PseudoEncoder.h"
+
+void set_up_encoder(PseudoEncoder *encoder, void (*isr_encoder)(void)) 
+{
+    pinMode(encoder->motor_pulse_pin, INPUT_PULLUP);
+    pinMode(encoder->channel_a_out, OUTPUT);
+    digitalWrite(encoder->channel_a_out, LOW);
+    attachInterrupt(digitalPinToInterrupt(encoder->motor_pulse_pin), isr_encoder, CHANGE);
```
noaht11: `isr_encoder` is attached to the `motor_pulse_pin`
Might be better to name it `isr_motor_pulse`?

PseudoGantry/.vscode/extensions.json
```diff
@@ -0,0 +1,7 @@
+{
+    // See http://go.microsoft.com/fwlink/?LinkId=827846
+    // for the documentation about the extensions.json format
+    "recommendations": [
+        "platformio.platformio-ide"
+    ]
+}
```
noaht11, line 7: Lets gitignore this file as well

noaht11 (changes requested):
Can you add the ability to not go 1-for-1 motor step to encoder count? (for 1-to-1, I can just bridge the motor pulse pin to the encoder pin on the Due).
I wanna test it with the actual ratio of like 5 encoder counts for every 8 motor steps (or some configurable ratio).

Also is limit switch support in progress? done? that's the other big thing I wanna test. Right now I'm just pushing a button manually which is not ideal.

If it's too difficult to develop without being able to test, I can try to implement those (or give you an Uno somehow), just let me know.

PseudoGantry/src/main.cpp
alexproskurin: Hahaha yeah

alexproskurin:
Limit switched should be good to go. In theory, should even support homing behavior.

alexproskurin:
Added Limit switch support and steps and counts specification.

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
```diff
@@ -0,0 +1,121 @@
...
+        if (pseudo_axis->motor_position_current < pseudo_axis->axis_length_counts)
+        {
+            pseudo_axis->motor_position_current++;
+            // check if limit switch was pressed before this move
+            // checks whether the homing routine is being run
+            if (!pseudo_axis->ls_home.status)
+            { // if !PRESSED where PRESSED = 0
+                // sets limit switch to unpressed
+                pseudo_axis->ls_home.status = UNRESSED;
+                // using digital write so it can work on Due ot Uno, execution time ~ 3 microseconds
```
chuandudx: "Due or Uno" :)

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
```diff
@@ -0,0 +1,121 @@
...
+    else if (!direction)
+    {
+        if (pseudo_axis->motor_position_current > 0)
+        {
+            pseudo_axis->motor_position_current--;
+
+            if (!pseudo_axis->ls_far.status)
+            { // if !PRESSED where PRESSED = 0
+                // sets limit switch to unpressed
+                pseudo_axis->ls_far.status = UNRESSED;
```
chuandudx: Typo `UNRESSED`

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+                digitalWrite(pseudo_axis->ls_far.output_pin, UNRESSED);
```
chuandudx: Typo `UNRESSED`

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
```diff
@@ -0,0 +1,121 @@
...
+    pseudo_axis->motor_position_current = pseudo_axis->motor_position_default;
+}
+
+void dump_data(PseudoAxis *pseudo_axis)
+{
+    if (Serial) {
+        String home_status;
+        String far_status;
+        Serial.print("Axis:              "); Serial.println(pseudo_axis->axis_name);
+        Serial.print("Axis length:       "); Serial.println(pseudo_axis->axis_length_counts);
```
chuandudx: Can we add units to the print statement? ie. `"Axis length (counts): "` and `"Motor position (units)"`

PseudoGantry/lib/PseudoGantry/PseudoAxis.h
```diff
@@ -0,0 +1,45 @@
...
+{   
+    String axis_name;
+    PseudoEncoder encoder;
+    uint32_t axis_length_counts;
+    uint32_t motor_position_current;
+    uint32_t motor_position_default;
+    uint32_t motor_dir_pin;
+    volatile int skip_counter;
+    int steps_for_ratio;                // values necesarry to determine steps vs counts ration
+    int counts_for_ratio;               // values necesarry to determine steps vs counts ration
```
chuandudx: ratio :) sorry for being nitpicky, just noting typos in the comments as well so they can be fixed for better readability.

PseudoGantry/src/main.cpp
```diff
@@ -0,0 +1,143 @@
+#include <Arduino.h>
+#include "PseudoAxis.h"
+
+// pins and constants for testing pseudo gantry
+#define PWM_FREQ        245.
+#define PWM_PIN         11.
+#define STEPS_TO_TAKE   800.
+float delay_millis = ((1000/PWM_FREQ/2)*STEPS_TO_TAKE);
```
chuandudx: I think the numbers all have to be floats to avoid integer truncation? ie `(1000.0/PWM_FREQ/2.0)*STEPS_TO_TAKE`

alexproskurin: I tested and wasn't getting any integer truncation, but I'll change them for consistency

PseudoGantry/lib/PseudoGantry/PseudoAxis.h
noaht11: Also `necesarry` --> `necessary`

PseudoGantry/src/main.cpp
```diff
@@ -0,0 +1,143 @@
...
+  //     .changes_to_skip = 2 * (STEPS - COUNTS),
+  //     .ls_home = {
+  //       .output_pin = LIMIT_SW_OUT_HOME_Y,
+  //       .status = UNRESSED
+  //     },
+  //     .ls_far = {
+  //       .output_pin = LIMIT_SW_OUT_FAR_Y,
+  //       .status = UNRESSED
+  //     }
+  //     };
```
noaht11: How come y is commented out?

PseudoGantry/lib/PseudoGantry/PseudoAxis.h
```diff
@@ -0,0 +1,45 @@
+#ifndef PSEUDO_AXIS_H
+#define PSEUDO_AXIS_H
+
+#include "Arduino.h"
+
+typedef struct PseudoEncoder
+{
+    uint32_t motor_pulse_pin;
+    volatile uint32_t channel_a_out;
```
noaht11: This is a pin number right? So shouldn't need to be volatile. Also might wanna call it `channel_a_pin` to make it clear

PseudoGantry/lib/PseudoGantry/PseudoAxis.h
```diff
@@ -0,0 +1,45 @@
...
+} PseudoLimitSwitch;
+
+typedef enum {
+    PRESSED = LOW,
+    UNRESSED = HIGH
+} LimitSwitchStatus;
+
+typedef struct PseudoAxis
+{   
+    String axis_name;
```
noaht11: I would just use a single `char` here, String seems unnecessary

PseudoGantry/src/main.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+#define LIMIT_SW_OUT_HOME_Y 10
+#define LIMIT_SW_OUT_FAR_Y 11
+
+// steps to counts constants
+// only supports a mode where number of steps > number of counts
+#define STEPS 8
+#define COUNTS 5
+
+// gentry length definition in counts
```
noaht11: Typo: `gantry`

PseudoGantry/src/main.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+void loop()
+{
+  // put your main code here, to run repeatedly:
+  dump_data(&pseudo_axis_x);
+  delay(5000);
+  analogWrite(PWM_PIN, 122);
+  delay(delay_millis);
+  analogWrite(PWM_PIN, 0);
+}
```
noaht11, line 133: nit: `setup` and `loop` are indented 2 spaces while everything else is 4

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+void reset_pseudo_axis(PseudoAxis *pseudo_axis)
+{
+    pseudo_axis->skip_counter=0;
+    pseudo_axis->motor_position_current = pseudo_axis->motor_position_default;
+}
+
+void dump_data(PseudoAxis *pseudo_axis)
+{
+    if (Serial) {
```
noaht11, line 102: What's this check for? Serial will always be defined on any Arduino board I can think of

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
```diff
@@ -0,0 +1,121 @@
...
+            pseudo_axis->ls_home.status = PRESSED;
+            digitalWrite(pseudo_axis->ls_home.output_pin, PRESSED);
+        }
+    }
+    // dump_data(pseudo_axis);
+}
+
+void reset_pseudo_axis(PseudoAxis *pseudo_axis)
+{
+    pseudo_axis->skip_counter=0;
```
noaht11: nit: spaces around `=`

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
```diff
@@ -0,0 +1,121 @@
...
+    pseudo_axis->motor_position_current = pseudo_axis->motor_position_default;
```
noaht11, line 95: Should you reset the limit switch statuses here as well?

PseudoGantry/lib/PseudoGantry/PseudoAxis.h
```diff
@@ -0,0 +1,45 @@
...
+    PseudoEncoder encoder;
+    uint32_t axis_length_counts;
+    uint32_t motor_position_current;
+    uint32_t motor_position_default;
+    uint32_t motor_dir_pin;
+    volatile int skip_counter;
```
noaht11: I think this would be better named `edge_counter` or `change_counter`, since it's not just counting the number of edges that are skipped it counts every edge until the full step count is reached

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+bool toggle_encoder_output(PseudoAxis *pseudo_axis)
+{
+    pseudo_axis->skip_counter++;
+    // skip steps in the beginning
+    if (pseudo_axis->skip_counter <= pseudo_axis->changes_to_skip) return false;
+    // else start toggling until the next set of STEPS
+    else if ((pseudo_axis->skip_counter <= (2 * pseudo_axis->steps_for_ratio)) || (pseudo_axis->changes_to_skip == 0))
+    {   
+        bool motor_pin_status = digitalRead(pseudo_axis->encoder.motor_pulse_pin);
```
noaht11: `digitalRead` returns an `int` so I would declare this as an `int` or explicitly cast it

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
```diff
@@ -0,0 +1,121 @@
...
+        // Serial.print("motor pin:    "); Serial.println(motor_pin_status);
```
noaht11: Remove?

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
```diff
@@ -0,0 +1,121 @@
...
+        // check if encoder count needs to be incremented
+        // if HIGH it means triggered on rising
+        if (motor_pin_status) return true;
+        else return false;
+    }
+    else
+    {
+        // this is assigned 1 and not 0 because at this point our step_count_x is ((2 * STEPS) + 1)
+        // which means we are onto the next set of 8, and 17 would be the beginnig of that set
+        pseudo_axis->skip_counter = 1;
```
noaht11: If `steps_for_ratio` and `counts_for_ratio` are the same, then `changes_to_skip` will be `0`.
In that case we will never reach this statement (since the `else if` above will always evaluate to true). This means the `skip_counter` will count up until it wraps around to negative numbers (since it's declared as an `int`). Once it reaches negative numbers, the first `if (pseudo_axis->skip_counter <= pseudo_axis->changes_to_skip)` will start evaluating to `true` and we'll stop toggling the encoder output pin.
The problem wouldn't appear until we reach 2 billion counts, but if it's not too complicated to handle correctly I think it's worth doing.

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
```diff
@@ -0,0 +1,121 @@
...
+{
+    pseudo_axis->skip_counter++;
+    // skip steps in the beginning
+    if (pseudo_axis->skip_counter <= pseudo_axis->changes_to_skip) return false;
+    // else start toggling until the next set of STEPS
+    else if ((pseudo_axis->skip_counter <= (2 * pseudo_axis->steps_for_ratio)) || (pseudo_axis->changes_to_skip == 0))
+    {   
+        bool motor_pin_status = digitalRead(pseudo_axis->encoder.motor_pulse_pin);
+        // Serial.print("motor pin:    "); Serial.println(motor_pin_status);
+        digitalWrite((pseudo_axis->encoder).channel_a_out, motor_pin_status ^ HIGH);
```
noaht11: Based on this, the encoder output is the inversion of the motor pulse. In the actual firmware, we increment our encoder count on the `RISING` edge of the encoder. Here you're returning true (and then incrementing the encoder count) on the rising edge of the motor therefore the `FALLING` edge of the encoder output.
There's nothing necessarily wrong with that (it just means the pseudo gantry will update it's encoder count a pulse width earlier than the actual firmware). Just wanna make sure that was intentional and you were aware of it.

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
```diff
@@ -0,0 +1,121 @@
...
+        // check if encoder count needs to be incremented
+        // if HIGH it means triggered on rising
+        if (motor_pin_status) return true;
+        else return false;
```
noaht11: This can be simplified to `return (motor_pin_status == HIGH);`

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+    if (direction)
+    {
+        if (pseudo_axis->motor_position_current < pseudo_axis->axis_length_counts)
+        {
+            pseudo_axis->motor_position_current++;
+            // check if limit switch was pressed before this move
+            // checks whether the homing routine is being run
+            if (!pseudo_axis->ls_home.status)
+            { // if !PRESSED where PRESSED = 0
```
noaht11: Would be clearer to remove the comment and write `if (pseudo_axis->ls_home.status == PRESSED)`

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+    else if (!direction)
+    {
+        if (pseudo_axis->motor_position_current > 0)
+        {
+            pseudo_axis->motor_position_current--;
+
+            if (!pseudo_axis->ls_far.status)
+            { // if !PRESSED where PRESSED = 0
```
noaht11: Would be clearer to remove the comment and write `if (pseudo_axis->ls_far.status == PRESSED)`

PseudoGantry/src/main.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+void loop()
+{
+  // put your main code here, to run repeatedly:
+  dump_data(&pseudo_axis_x);
+  delay(5000);
+  analogWrite(PWM_PIN, 122);
+  delay(delay_millis);
+  analogWrite(PWM_PIN, 0);
```
noaht11: I assume this was for your "acting as a motor" testing?
Can probably replace this with just:
```c
dump_data(&pseudo_axis_x);
dump_data(&pseudo_axis_y);
delay(1000);
```
or something similar

noaht11 (changes requested):

PseudoGantry/lib/PseudoGantry/PseudoAxis.h
```diff
@@ -0,0 +1,45 @@
...
+typedef struct PseudoEncoder
+{
+    uint32_t motor_pulse_pin;
+    volatile uint32_t channel_a_out;
+} PseudoEncoder;
+
+typedef struct PseudoLimitSwitch
+{
+    uint32_t output_pin;
+    uint32_t status;
```
noaht11: Can declare this as a `LimitSwitchStatus` (move the enum above this)

PseudoGantry/src/main.cpp
alexproskurin: was just testing one axis

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
alexproskurin: Actually, that was not my intention, so i will change it to: `digitalWrite((pseudo_axis->encoder).channel_a_pin, motor_pin_status);`

PseudoGantry/lib/PseudoGantry/PseudoAxis.h
```diff
@@ -0,0 +1,45 @@
+#ifndef PSEUDO_AXIS_H
+#define PSEUDO_AXIS_H
+
+#include "Arduino.h"
+
+typedef struct PseudoEncoder
+{
+    uint32_t motor_pulse_pin;
+    volatile uint32_t channel_a_pin;
```
noaht11: Still doesn't need to be volatile right?

noaht11 (changes requested):

PseudoGantry/src/main.cpp
alexproskurin: for some reason, main setting was set to 2 lol

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
alexproskurin: On UNO if (Serial) returns true if USB serial connection is open with a computer

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
alexproskurin: damn, that is a really good catch!!! I just changed it to `if (pseudo_axis->changes_to_skip != 0) pseudo_axis->change_counter++;`

PseudoGantry/lib/PseudoGantry/PseudoAxis.h
alexproskurin: i dont know why i ever set it to volatile LOL

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
noaht11: I believe that almost fixes it, but `change_counter` starts set to 0, which means if `changes_to_skip` is 0, it'll never be incremented.
Then `if (pseudo_axis->change_counter <= pseudo_axis->changes_to_skip)` will always evaluate to true (since both values are 0).

You could fix it by setting `change_counter` to 1 in `reset_pseudo_axis` only if `changes_to_skip` is 0 although that's kinda hacky.

Better approach might be to replace:
```c
if (pseudo_axis->change_counter <= pseudo_axis->changes_to_skip)
```
with
```c
if ((pseudo_axis->change_counter <= pseudo_axis->changes_to_skip) && (pseudo_axis->changes_to_skip != 0))
```

PseudoGantry/lib/PseudoGantry/PseudoAxis.h
noaht11: Ya never know, the pins might spontaneously re-arrange on you :)

PseudoGantry/lib/PseudoGantry/PseudoAxis.cpp
alexproskurin: Good idea! Fixed

PR #16, opened 2020-04-01
Motor Refactor

noaht11:
**General objectives of this PR**
- Extracting configuration to the top-level
- Making more use of hardware
- Removing duplicated code
- General cleanup and documentation

**Specific changes**
- Deleted LimitSwitch.h and Encoder.h and moved their relevant pieces into Axis
- The step PWM signal is now generated in hardware by the timer `TIOA` output. This allows an easily configurable duty cycle and reduces the requirements for the velocity interrupt
- The velocity interrupt now just updates the counter reset value (i.e. the frequency) of the timer for the step signal
- Debouncing the limit switches in now handled in hardware (a feature of the Due PIO)
- Extracted all the duplicated code between the axes into common inline functions
- Added validation checks to `start_axis` to check all the required conditions are met before starting to move an axis
- The homing routine is now handled by the top-level rather than within Axis to keep the scope of Axis down and eliminate cyclic dependencies. This also simplifies the limit switch ISRs so now they just stop movement and update a state variable
- Deconstructed the `Axis` struct into several smaller structs with well-defined purposes.
- The X and Y instances of the `Axis` struct are now internal to Axis.cpp and external access is accomplished using `AxisId` values
- Configuration of which pins are used by the axes is now controllable from the top-level via the `AxisIO` struct
- The movement information passed to an axis is now overspecified since it includes the holding velocity; however, the axis will stop accelerating once it reaches the holding velocity minimizing the accumulated encoder count error (since we will be guaranteed to be holding at the correct velocity)
- Constrained `Kinematics` to operate in pure math, independent of specific units. Unit conversion (steps to/from counts etc.) are now handled in `Movement`
- Some default movement parameters and encoder count / motor step specifications are now configurable from the top-level via the `mPMTTestStandMotionConfig` struct
- Added periodic dump of axis state information to the debug serial
- Cleaned up top-level state machine handling of axis states
- Added comment documentation to almost all of the Gantry modules


MessageTerminal/MessageTerminal.cxx
```diff
@@ -99,7 +99,6 @@ void get_status(istringstream& iss)
                 case STATUS_IDLE:   puts("Status: IDLE"); break;
                 case STATUS_MOVING: puts("Status: MOVING"); break;
                 case STATUS_HOMING: puts("Status: HOMING"); break;
-                case STATUS_FAULT:  puts("Status: FAULT"); break;
                 default:            puts("ERR: Invalid status"); break;
```
chuandudx, line 102: What's the reason for removing the fault status?

firmware/lib/Gantry/Axis.cpp
```diff
@@ -1,311 +1,569 @@
...
+/** TC IRQ number for y-axis velocity timer */
+#define IRQ_Y_AXIS_VELOCITY    0
+/** TC IRQ number for y-axis acceleration timer */
+#define IRQ_Y_AXIS_ACCEL       1
+
+/** Frequency of the slow clock */
+#define SCLK_FREQ              32768
+
+/** Minimum pulse duration to pass debouncing */
+#define DEBOUNCE_FILTER_MS     50
```
chuandudx, line 24: Is this with regards to the encoder? We are waiting 50 milliseconds for what action?

firmware/lib/Gantry/Axis.cpp
```diff
@@ -1,311 +1,569 @@
...
+
+/** Macro to determine if we've reached a target encoder count based on direction */
+#define REACHED_TARGET(_dir, _cur, _tar)             \
+    (((_dir) == DIR_POSITIVE && (_cur) >= (_tar)) || \
+    ((_dir) == DIR_NEGATIVE && (_cur) <= (_tar)))
+
```
chuandudx, line 30: maximum macro-tude

firmware/lib/Gantry/Axis.cpp
```diff
@@ -1,311 +1,569 @@
...
+/** Maximum allowed velocity for an axis */
+#define VEL_MAX                25000
```
chuandudx, line 32: Can we add its units in comment or in the variable name for clarity. In this case, it looks like counts?

firmware/lib/Gantry/Axis.h
```diff
@@ -1,69 +1,108 @@
...
-void stop_axis(Axis *axis);
+/**
+ * @enum AxisResult
+ * 
+ * @brief Possible return values for a call to axis_start
+ */
+typedef enum {
+    AXIS_OK,                           //!< Axis movement started OK
+    AXIS_ERR_DIR_MISMATCH,             //!< Direction did not match count value signs
+    AXIS_ERR_ZERO_DIST,                //!< Total specified distance was zero
```
chuandudx: Does specifying zero distance qualify as an error? Or should we just not move that axis.

firmware/lib/Gantry/Kinematics.h
```diff
@@ -4,12 +4,12 @@
 #include <stdint.h>
 
 typedef struct {
-    uint32_t counts_accel;
-    uint32_t counts_hold;
-    uint32_t counts_decel;
+    uint32_t dist_accel;
```
chuandudx: Similar comment - can we clarify the unit of `dist` (counts or mm)?

firmware/lib/Gantry/Movement.cpp
```diff
@@ -1,54 +1,64 @@
...
+ * 
+ * @param axis_id       The AxisId identifying the axis
+ * @param dir           The direction to move (either DIR_POSITIVE or DIR_NEGATIVE)
+ * @param total_counts  The total distance [encoder counts]
+ * @param shape         Pointer to a MotionShape struct specifying the derivatives of the motion
+ * @param units         Pointer to a MotionUnits struct specifying the conversion ratios between
+ *                      encoder counts and motor steps
+ * 
+ * @return true if the axis successfully started executing the motion, otherwise false
+ */
```
chuandudx: nice description!

firmware/lib/Gantry/Movement.cpp
```diff
@@ -1,54 +1,64 @@
...
-        return;
-    }
-}
\ No newline at end of file
+    return move_axis_rel(axis_id, DIR_NEGATIVE, INT32_MAX, shape, units);
+}
+
+bool move_axis_home_b(AxisId axis_id, MotionShape shape, MotionUnits units)
+{
+    return move_axis_rel(axis_id, DIR_POSITIVE, INT32_MAX, shape, units);
```
chuandudx: This is always specifying a max possible number of count to move, until the limit switch is reached?

firmware/lib/Gantry/Timer.cpp
```diff
@@ -1,56 +1,151 @@
...
+ * 
+ * | Name         | Definition |
+ * | ------------ | ---------- |
+ * | TIMER_CLOCK1 | MCK/2      |
+ * | TIMER_CLOCK2 | MCK/8      |
+ * | TIMER_CLOCK3 | MCK/32     |
+ * | TIMER_CLOCK4 | MCK/128    |
+ * | TIMER_CLOCK5 | SLCK       |
+ */
+
```
chuandudx, line 32: Nice addition!

firmware/lib/Gantry/Timer.h
```diff
@@ -1,14 +1,24 @@
...
+#define TC_IRQN(_x)            TC_IRQN_I(_x)
 
-void reset_timer(Tc *tc, uint32_t channel, IRQn_Type irq, uint32_t velocity);
+#define TC_ISR_I(_x)           void TC##_x##_Handler()
+/** Generates the ISR function name corresponding to a TC IRQ number */
+#define TC_ISR(_x)             TC_ISR_I(_x)
 
-void start_timer_accel(Tc *tc, uint32_t channel, IRQn_Type irq, uint32_t accel);
+// PWM Timers
+void configure_pwm_timer(Tc *tc, uint32_t channel, IRQn_Type irq, Pio *pio_bank, EPioType periph, uint32_t pin_mask);
```
chuandudx, line 15: Just realizing that all the functions use underscore convention as opposed to camelCase?

firmware/src/mPMTTestStand.cpp
```diff
@@ -111,34 +112,108 @@ void mPMTTestStand::handle_get_data()
...
+    switch (this->status) {
+        case STATUS_IDLE:
+            break;
+        case STATUS_MOVING:
+            if (this->x_state->moving || this->y_state->moving) {
+                this->status = STATUS_MOVING;
+            }
+            else if (this->x_state->ls_home_pressed || this->x_state->ls_far_pressed ||
+                     this->y_state->ls_home_pressed || this->x_state->ls_far_pressed) {
+                this->status = STATUS_FAULT;
```
chuandudx, line 187: Did we remove the fault status earlier on?

MessageTerminal/MessageTerminal.cxx
noaht11: Oops I removed it then added it back everywhere except here. Will add it back here as well.

firmware/lib/Gantry/Axis.cpp
noaht11: This is for debouncing the limit switches. The limit switch value has to stay steady for 50 ms to be accepted

firmware/lib/Gantry/Axis.cpp
noaht11: It's actually steps / second here, so evidently good point. I actually realize I need to move this up a couple levels, so will do that and add units in a comment

firmware/lib/Gantry/Axis.h
noaht11: I guess there's nothing wrong with accepting a zero distance movement. I can switch it to not be an error

firmware/lib/Gantry/Kinematics.h
noaht11: Ahh, it's described in the comments on `generate_vel_profile` in `Kinematics.cpp`. The units are actually intentionally not specified because it's up to the user. I made it so this `Kinematics` module deals in pure math where it'll use whatever units are used for parameters that get passed to functions.
I can add some comments here to clarify that.

firmware/lib/Gantry/Movement.cpp
noaht11: Yeah, the idea is try to move forever, the limit switch will automatically stop you once you've reached the home.

firmware/lib/Gantry/Timer.h
noaht11: Most of the C code I've ever dealt with uses `snake_case`.
Idk what the C++ convention is, but all of the Axis code is actually pure C so I think we should stick with snake_case here.

firmware/src/mPMTTestStand.cpp
noaht11: It was removed, then re-added, but I forgot it in that other spot, see the comment above.

firmware/lib/Gantry/Kinematics.cpp
```diff
@@ -1,38 +1,83 @@
...
+//  * 
+//  * @param accel          The acceleration rate [motor steps / s^2]
+//  * @param v_min          The starting velocity [motor steps / s]
+//  * @param elapsed_counts The elapsed number of encoder counts
+//  */
+// uint32_t calc_expected_velocity(uint32_t accel, uint32_t v_min, uint32_t elapsed_counts)
+// {
+//     uint32_t elapsed_steps = elapsed_counts * MOTOR_STEPS_PER_REV / ENCODER_COUNTS_PER_REV;
+//     return (uint32_t)sqrt((2 * accel * elapsed_steps) + (v_min * v_min));
+// }
```
noaht11: Need to delete this

firmware/lib/Gantry/Kinematics.h
```diff
@@ -4,12 +4,12 @@
...
+    uint32_t dist_hold;
+    uint32_t dist_decel;
 } VelProfile;
 
-uint32_t calc_counts_accel(uint32_t accel, uint32_t v_min, uint32_t v_max);
-bool generate_vel_profile(uint32_t accel, uint32_t v_min, uint32_t v_max, uint32_t counts_total, VelProfile *profile_out);
+// uint32_t calc_expected_velocity(uint32_t accel, uint32_t v_min, uint32_t elapsed_counts);
```
noaht11: Need to delete

chuandudx:
As a general question for the limit switches, are they already initialized with `INPUT_PULLUP`. If so, I will exclude a hardware resistor pull-up from schematic to simplify the soldering job required.

firmware/lib/Gantry/Axis.cpp
```diff
@@ -1,311 +1,569 @@
...
-#define LIMIT_SW_FAR_PIN_Y      27
-
-// Interrupt Declarations
 void isr_encoder_x();
-void isr_limit_switch_x();
+void isr_ls_home_x();
+void isr_ls_far_x();
+
+static Axis axis_x = {
+    .io = {},
```
alexproskurin, line 98: Didn't know you could do this dot thing, pretty sweet!

noaht11:
> As a general question for the limit switches, are they already initialized with `INPUT_PULLUP`. If so, I will exclude a hardware resistor pull-up from schematic to simplify the soldering job required.

@chuandudx Yeah they're internally pulled up so do not need an external pull-up

firmware/lib/Gantry/Axis.cpp
noaht11: Yup, great way to initialize structs!

chuandudx:
consider removing pin assignments `ENCODER_PIN_B_X` and `ENCODER_PIN_B_Y` since they aren't used.

firmware/lib/Gantry/Axis.cpp
```diff
@@ -1,311 +1,569 @@
...
+ * 
+ * @param axis Pointer to the Axis to use
+ */
+static void setup_interrupts(Axis *axis)
 {
-    attachInterrupt(digitalPinToInterrupt(axis_config->encoder_pin_a), axis_config->isr_encoder, RISING);
-    delay(1000);
-    attachInterrupt(digitalPinToInterrupt(axis_config->ls_home), axis_config->isr_limit_switch, CHANGE);
-    delay(1000);
-    attachInterrupt(digitalPinToInterrupt(axis_config->ls_far), axis_config->isr_limit_switch, CHANGE);
```
alexproskurin, line 82: I don't think this one should be set up to trigger on `CHANGE`. I think we should only trigger it when it is first pressed, so `FALLING`. I think the current set up breaks the homing routine in a case when we have `ls_far` `PRESSED` and we start homing.

firmware/lib/Gantry/Axis.cpp
```diff
@@ -1,311 +1,569 @@
...
-#include "Timer.h"
 #include "Movement.h"
+#include "Timer.h"
+
+/*****************************************************************************/
+/*                                  DEFINES                                  */
+/*****************************************************************************/
+
+/** TC IRQ number for x-axis velocity timer */
+#define IRQ_X_AXIS_VELOCITY    3
```
alexproskurin, line 11: Kinda nitpicking here, I think a better naming convention is: `IRQ_AXIS_VELOCITY_X` for X and `IRQ_AXIS_VELOCITY_Y` for Y, that is having a differentiating identifier at the end of the variable name.

firmware/lib/Gantry/Axis.cpp
```diff
@@ -1,311 +1,569 @@
...
+    setup_debouncing(axis->io.pin_ls_home, DEBOUNCE_FILTER_MS);
+    attachInterrupt(digitalPinToInterrupt(axis->io.pin_ls_home),
+                    axis->interrupts.isr_ls_home,
+                    CHANGE);
+
+    // Debounce and attach interrupt to FAR limit switch
+    setup_debouncing(axis->io.pin_ls_far, DEBOUNCE_FILTER_MS);
+    attachInterrupt(digitalPinToInterrupt(axis->io.pin_ls_far),
+                    axis->interrupts.isr_ls_far,
+                    CHANGE);
```
alexproskurin: I think i commented on already removed code for this, but still the same thing - switch `ls_far` from `CHANGE` to `FALLING`

firmware/lib/Gantry/Axis.cpp
```diff
@@ -1,311 +1,569 @@
...
+
+    // Reject if we've hit the far limit switch and are trying to move forward
+    if (motion->dir == DIR_POSITIVE && axis->state.ls_far_pressed) return AXIS_ERR_LS_FAR;
+    // Reject if we've hit the home limit switch and are trying to move backward
+    if (motion->dir == DIR_NEGATIVE && axis->state.ls_home_pressed) return AXIS_ERR_LS_HOME;
+
+    // Configure state
+    axis->state.moving = true;
+    axis->state.velocity = motion->vel_start;
+    axis->state.next_velocity = axis->state.velocity;
```
alexproskurin, line 285: What is the next velocity for?

firmware/lib/Gantry/Axis.cpp
```diff
@@ -1,311 +1,569 @@
...
+    // Start velocity PWM timer
+    reset_pwm_timer(
+        axis->io.tc_step,
+        axis->io.tc_step_channel,
+        axis->io.tc_step_irq,
+        axis->state.velocity,
+        VEL_DUTY_CYCLE);
+
+    // Start acceleration timer interrupt
+    reset_timer_interrupt(
```
alexproskurin, line 302: Do we only use this for acceleration? Might be more informative to just have it be named `reset_acceleration_timer_interrupt`?

firmware/lib/Gantry/Axis.cpp
```diff
@@ -1,311 +1,569 @@
...
-            }
-        }
-        else {
-            axis_x.fault = true;
-        }
-    }
-    last_interrupt_time = interrupt_time;
+    stop_axis(axis);
+    // Limit switch reads LOW when pressed
+    axis->state.ls_home_pressed = (digitalRead(axis->io.pin_ls_home) == LOW);
```
alexproskurin: Instead of comparing to `LOW`, would it be easier to read if it was compared to limit switch status enum `PRESSED`. That way if the limit switch wiring ever changes you could just swap the order of limit switch status assignment in enum? Orr is this not very C like?

firmware/lib/Gantry/Axis.cpp
```diff
@@ -1,311 +1,569 @@
...
+/**
+ * @brief Common FAR limit switch ISR handler
+ * 
+ * @param axis Pointer to the Axis whose limit switch triggered the interrupt
+ */
+static __attribute__((always_inline)) inline void handle_isr_ls_far(Axis *axis)
+{
+    stop_axis(axis);
+    // Limit switch reads LOW when pressed
+    axis->state.ls_far_pressed = (digitalRead(axis->io.pin_ls_far) == LOW);
```
alexproskurin: Same as above

firmware/lib/Gantry/Axis.cpp
```diff
@@ -1,311 +1,569 @@
...
+ * @brief Common acceleration timer ISR handler
+ * 
+ * @param axis Pointer to the Axis whose acceleration timer triggered the interrupt
+ */
+static __attribute__((always_inline)) inline void handle_isr_accel(Axis *axis)
+{
+    if (!axis->state.moving) return;
+
+    switch (axis->state.velocity_segment) {
+        case VEL_SEG_ACCELERATE:
```
alexproskurin, line 427: I think i was reading somewhere that if you have less than 5 switch cases, simple if statements could be faster

firmware/lib/Gantry/Axis.cpp
```diff
@@ -1,311 +1,569 @@
...
+                }
+            }
+            else {
+                // Stop moving once we reach the final target encoder count
+                stop_axis(axis);
+            }
+            break;
         }
     }
 }
```
alexproskurin, line 471: Not really sure if we can achieve the same accuracy without checking encoder counts and just relying on specified velocity. Just seems a little dangerous not to have a secondary check, but i might be missing something that got changed elsewhere

firmware/lib/Gantry/Axis.cpp
```diff
@@ -1,311 +1,569 @@
...
+/*****************************************************************************/
+/*                                 TYPEDEFS                                  */
+/*****************************************************************************/
+
+/**
+ * @struct AxisInterrupts
+ * 
+ * @brief Specifies all of the internal interrupts and ISRs used by the axis
+ */
+typedef struct {
```
alexproskurin, line 59: How come are these defined in .cpp and not in .h?

firmware/lib/Gantry/Axis.cpp
noaht11: Next velocity is so that if the velocity hasn't changed (e.g. during the holding segment or if we're accelerating slowly compare to our current velocity) we can notice that and not bother resetting the timer in the velocity ISR

firmware/lib/Gantry/Axis.cpp
noaht11: switch statements will definitely be faster for 5+ items because they will get compiled into a jump table. For less than 5 items there is negligible difference, so I went with switch cause it seemed more legible

firmware/lib/Gantry/Axis.cpp
noaht11: They are not used by (and should not be visible to) anyone outside of this .cpp file

firmware/lib/Gantry/Axis.cpp
noaht11: There's nothing in the function itself that is inherent to acceleration. It starts a timer interrupt at some frequency, could be used for anything

firmware/lib/Gantry/Axis.cpp
noaht11: I like it (as in your suggestion)

firmware/lib/Gantry/Axis.cpp
noaht11: There's nothing in the function itself that is inherent to acceleration. It starts a timer interrupt at some frequency, could be used for anything

firmware/lib/Gantry/Axis.cpp
noaht11: Good idea

firmware/lib/Gantry/Axis.cpp
noaht11: Sounds good

firmware/lib/Gantry/Axis.cpp
noaht11: You mean for the accelerate case? Both hold and decelerate check the encoder counts.
At the end of accelerate, I still have your error adjustment so if we reached holding velocity early, we'll hold for longer or if we reached it late, we'll hold for less time, so the accuracy should be the exact same as before, unless I'm missing something?

What this improves is we will never be holding at the wrong velocity. The final exit condition (i.e. the only way to reach stop_axis) is still to have reached to correct final encoder count.

firmware/lib/Gantry/Axis.cpp
noaht11: Actually I changed my mind. I picture the hierarchy as:

- Gantry
  - X Axis
    - Velocity Interrupt
    - Accel Interrupt
    - Other Interrupts...
  - Y Axis
    - Velocity Interrupt
    - Accel Interrupt
    - Other Interrupts...

So I'm narrowing down in order of hierarchy: first identify the axis, then which interrupt

Not opposed to switching to `IRQ_AXIS_X_VELOCITY` and `IRQ_AXIS_Y_VELOCITY` etc.

firmware/lib/Gantry/Axis.cpp
noaht11: Although like you mentioned in the other comment about switching the wiring, whether this is `FALLING` or `RISING` should be defined in the same place as whether pressed is `LOW` or `HIGH`

firmware/lib/Gantry/Axis.cpp
noaht11: The potential issue is if the hold_vel passed to start_axis doesn't match the counts. I think I'm actually going to eliminate the Movement files and put all that calculation logic into the Axis so it's not dependent on correct values being passed to it

firmware/src/main.cpp
```diff
@@ -1,35 +1,69 @@
...
+            .pin_dir            = 23, // PA14
+            .pin_enc_a          = 24,
+            .pin_enc_b          = 25,
+            .pin_ls_home        = 26,
+            .pin_ls_far         = 27
+        }
+    },
+    .gantry = {
+        .axis_mech = {
+            .counts_per_rev     = 800,    // For Igus gantry encoders
```
chuandudx: IGUS encoders are 500 counts/rev

firmware/src/main.cpp
```diff
@@ -1,35 +1,69 @@
...
+#include <Arduino.h>
+
+const mPMTTestStandConfig conf = {
+    .io = {
+        // Serial Devices
+        .serial_comm            = Serial,
+        .serial_comm_baud_rate  = 115200,
+        // Thermistor Pins
+        .pin_therm_amb          = A0,
+        .pin_therm_motor1       = A1,
```
chuandudx: Can we make `pin_therm_motor1` into `pin_therm_motor_axis_x` and  `pin_therm_motor2` into `pin_therm_motor_axis_y` . I think I was the one to call them motor1 and 2 but it's unclear what they refer to looking back.

firmware/lib/Gantry/Axis.cpp
noaht11: I made it so that the definition of the "pressed" state is part of the AxisIO, so we can configure it from there

firmware/src/main.cpp
noaht11: How about just `pin_therm_motor_x` and `pin_therm_motor_y` (no `axis` to keep it shorter)

chuandudx: sounds good to me!

noaht11:
> consider removing pin assignments `ENCODER_PIN_B_X` and `ENCODER_PIN_B_Y` since they aren't used.

@chuandudx I'm gonna leave them for now, we can purge them at the very end if we want

MessageTerminal/MessageTerminal.cxx
noaht11: Will address this in future PR

firmware/lib/Gantry/Axis.cpp
alexproskurin: Can we not define RISING or FALLING in terms of HIGH or LOW of the unpressed switch? I think internally RISING/FALLING maps to an integer.

firmware/lib/Gantry/Axis.cpp
alexproskurin: I could kinda see it go either way haha so i am good with either

firmware/lib/Gantry/Axis.cpp
alexproskurin: OK, I think we are good then

firmware/lib/Gantry/Axis.cpp
noaht11: I don't wanna depend on the internal definitions of RISING/FALLING (everything should still work if those definitions change).
Anyway now, the choice of RISING/FALLING is derived from the choice of LOW/HIGH for the pressed state

firmware/lib/Gantry/Axis.cpp
noaht11: Okay I'll go with my current one then to save the extra commit

PR #17, opened 2020-04-04
Serial Cleanup

noaht11:
* Adds documentation comments to all of the serial communication library
* Replaces true/false return values with enum return values to provide more specific error information

### DIFF
--- a/MessageTerminal/MessageTerminal.cxx
+++ b/MessageTerminal/MessageTerminal.cxx
@@ -7,6 +7,7 @@
 #include "TestStandCommHost.h"
 
 #include "shared_defs.h"
+#include "SerialResult.h"
 
 #include "macros.h"
 
@@ -17,12 +18,12 @@
 #define BASIC_CMD(_name)                     \
 void _name(istringstream& iss)               \
 {                                            \
-    if (comm._name()) {                      \
+    SerialResult res = comm._name();         \
+    if (res == SERIAL_OK) {                  \
         cout << "OK" << endl;                \
     }                                        \
     else {                                   \
-        printf("ERR: Received MSG ID: %d\n", \
-                comm.received_message().id); \
+        printf("ERROR: %d\n", res);          \
     }                                        \
 }
 
@@ -68,11 +69,12 @@ void move(istringstream& iss)
         else if (word == "neg") dir = DIR_NEGATIVE;
         else break;
 
-        if (comm.move(accel, hold_vel, dist, axis, dir)) {
+        SerialResult res = comm.move(accel, hold_vel, dist, axis, dir);
+        if (res == SERIAL_OK) {
             cout << "OK" << endl;
         }
         else {
-            printf("ERR: Received MSG ID: %d\n", comm.received_message().id);
+            printf("ERROR: %d\n", res);
             return;
         }
 
@@ -84,15 +86,17 @@ void move(istringstream& iss)
 
 void get_status(istringstream& iss)
 {
-    if (comm.get_status()) {
+    SerialResult res = comm.get_status();
+    if (res == SERIAL_OK) {
         cout << "OK" << endl;
     }
     else {
-        printf("ERR: Received MSG ID: %d\n", comm.received_message().id);
+        printf("ERROR: %d\n", res);
         return;
     }
 
-    if (comm.recv_message(RECV_MSG_TIMEOUT)) {
+    res = comm.recv_message(RECV_MSG_TIMEOUT);
+    if (res == SERIAL_OK) {
         if (comm.received_message().id == MSG_ID_STATUS) {
             Status status = (Status)((comm.received_message().data)[0]);
             switch (status) {
@@ -103,11 +107,11 @@ void get_status(istringstream& iss)
             }
         }
         else {
-            printf("ERR: Received MSG ID: %d\n", comm.received_message().id);
+            printf("ERROR: Wrong ID: %d\n", comm.received_message().id);
         }
     }
     else {
-        printf("ERR: No response\n");
+        printf("ERROR: %d\n", res);
     }
 }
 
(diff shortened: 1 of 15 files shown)


MessageTerminal/MessageTerminal.cxx
```diff
@@ -103,11 +107,11 @@ void get_status(istringstream& iss)
             }
```
chuandudx, line 107: Should we have `STATUS_FAULT`?

MessageTerminal/MessageTerminal.cxx
```diff
@@ -17,12 +18,12 @@
...
 {                                            \
-    if (comm._name()) {                      \
+    SerialResult res = comm._name();         \
+    if (res == SERIAL_OK) {                  \
         cout << "OK" << endl;                \
     }                                        \
     else {                                   \
-        printf("ERR: Received MSG ID: %d\n", \
-                comm.received_message().id); \
+        printf("ERROR: %d\n", res);          \
```
chuandudx, line 26: Adding `SerialResult.h` and printing the `res` makes it really clear!

firmware/src/mPMTTestStand.cpp
```diff
@@ -224,7 +224,7 @@ void mPMTTestStand::execute()
     }
 
     // Check for any messages
-    if (this->comm.check_for_message()) {
+    if (this->comm.check_for_message() == SERIAL_OK) {
         uint8_t id = this->comm.received_message().id;
         DEBUG_PRINT_VAL("Received Message w/ ID", id);
         switch (id) {
```
chuandudx, line 242: How does `handle_home_a()` relate to the axes. Similar for `handle_home_b()`?

shared/TestStandComm/SerialDevice.h
```diff
@@ -3,16 +3,67 @@
...
 
+/**
+ * @class SerialDevice
+ * 
+ * @brief Abstract base class representing an interface to serial communication hardware
+ *        on an arbitrary system
+ * 
+ * This is the lowest layer in the protocol stack and is only responsible for sending
+ * bytes. It performs no interpretation of the data being sent.
+ */
```
chuandudx, line 14: Thank you for the nice descriptions!

shared/TestStandComm/SerialTransport.h
```diff
@@ -4,25 +4,33 @@
...
 {
     private:
+        typedef enum {
+            MSG_SEG_START    = 0,
+            MSG_SEG_ID       = 1,
+            MSG_SEG_LENGTH   = 2,
+            MSG_SEG_DATA     = 3,
+            MSG_SEG_CRC      = 4,
+            MSG_SEG_END      = 5
+        } MessageSegment;
```
chuandudx, line 25: Unsure if you had already mentioned in another PR if there was a specific reason that you wanted to keep the enumeration explicit by having 0,1,2,3? Otherwise I think
`typedef enum{
MSG_SEG_START, MSG_SEG_ID,....} MessageSegment;` would be ok?

shared/TestStandComm/SerialTransport.h
```diff
@@ -4,25 +4,33 @@
...
+
+        typedef struct {
+            MessageSegment current_segment;
+            uint8_t bytes_read;
+            uint8_t msg_length;
```
chuandudx, line 30: I remember with the motor counts values, it had to be 16 bits to prevent the values from wrapping around. I know you fixed it somewhere but just wanted to bring up in case `msg_length` needs to be longer.

shared/TestStandComm/TestStandComm.cxx
```diff
@@ -1,11 +1,28 @@
...
 TestStandComm::TestStandComm(SerialDevice& device) : device(device), transport(device), session(transport, this->received_msg)
 {
     this->received_msg.data = this->received_data;
 }
 
-bool TestStandComm::send_basic_msg(uint8_t id)
+/**
+ * @brief Helper method to send a "basic message"
+ * 
+ * A "basic message" is a message with just an ID and no payload data.
```
chuandudx, line 19: sounds very orbit

firmware/src/mPMTTestStand.cpp
noaht11: "Homing" has two steps:
- A: Move the axes in the negative direction until you hit the limit switch
- B: Move the axes in the positive direction until the limit switch releases

shared/TestStandComm/SerialTransport.h
noaht11: Oh yeah forgot about that. No reason to keep them here

shared/TestStandComm/SerialTransport.h
noaht11: `msg_length` is a `uint8_t` meaning it can have any value between 0 and 255, meaning we can send a message up to 255 bytes long which is plenty (A motor move command sends like 14 bytes).

shared/TestStandComm/TestStandComm.cxx
noaht11: Whaaat, no, no orbit influence here at all :)

MessageTerminal/MessageTerminal.cxx
noaht11: Sorry forgot about this, it's coming up in the next PR

firmware/src/mPMTTestStand.cpp
chuandudx: Ah, that makes sense! Would it be possible to have a descriptions for this in future PR?

shared/TestStandComm/SerialTransport.h
chuandudx: Do you remember the spot where the size needed to be increased to accommodate for the encoder values? Just wanna check it out.

firmware/src/mPMTTestStand.cpp
noaht11: I can add it now, need to update this PR anyway

shared/TestStandComm/SerialTransport.h
noaht11: All of the motor-related numbers are 32-bit and have been updated everywhere except feArduino (but I updated that in my demo branch will I'll make a PR for soon)

MessageTerminal/MessageTerminal.cxx
```diff
@@ -7,6 +7,7 @@
 #include "TestStandCommHost.h"
 
 #include "shared_defs.h"
+#include "SerialResult.h"
```
alexproskurin, line 10: One is snake case, the other is camel, does this signify something?

shared/TestStandComm/SerialDevice.h
```diff
@@ -3,16 +3,67 @@
...
 {
     public:
+        /**
+         * @brief Open a connection to the serial hardware
+         * 
+         * @param baud_rate The baud rate for the serial connection
+         * 
+         * @return true if the connection opened successfully, otherwise false
+         */
         virtual bool ser_connect(uint32_t baud_rate) = 0;
```
alexproskurin, line 25: Since all of these are class methods, i would omit `ser_` in names

shared/TestStandComm/SerialSession.cxx
```diff
@@ -1,65 +1,130 @@
...
         .length = 0,
         .data = nullptr
     };
-    this->transport.send_message(ack);
+    return this->transport.send_message(ack);
 }
 
-bool SerialSession::check_for_message()
+/**
+ * @brief Checks if the received message was expected and sends an ACK if it was
```
alexproskurin, line 31: Is `SERIAL_OK` an `ACK`? Since you say that we return an `ACK` on success, but actually return `SERIAL_OK`

noaht11: We **send** an ACK (like send it over serial). We **return** `SERIAL_OK`

shared/TestStandComm/SerialDevice.h
noaht11: I would like to, but the problem with that is in the `LinuxSerialDevice` I have to call C standard library functions to actually talk to the hardware and they are called `open`, `read`, `write` etc. so it becomes ambiguous which one I'm calling

MessageTerminal/MessageTerminal.cxx
noaht11: Uhh, poor consistency?
I can fix that

noaht11: Actually, I'm gonna change it in a future PR cause otherwise I'll create conflicts with myself

PR #18, opened 2020-04-04
Simplified Serial Communication

noaht11:
- Integrated response handling into TestStandCommHost for `GET_STATUS`
- Replaced `GET_DATA` with 3 new messages:
  - `GET_POSITION`
  - `GET_AXIS_STATE`
  - `GET_TEMP`
- Implemented `GET_POSITION` (including response handling)

### DIFF
--- a/MessageTerminal/MessageTerminal.cxx
+++ b/MessageTerminal/MessageTerminal.cxx
@@ -86,28 +86,15 @@ void move(istringstream& iss)
 
 void get_status(istringstream& iss)
 {
-    SerialResult res = comm.get_status();
+    Status status;
+    SerialResult res = comm.get_status(&status, RECV_MSG_TIMEOUT);
     if (res == SERIAL_OK) {
-        cout << "OK" << endl;
-    }
-    else {
-        printf("ERROR: %d\n", res);
-        return;
-    }
-
-    res = comm.recv_message(RECV_MSG_TIMEOUT);
-    if (res == SERIAL_OK) {
-        if (comm.received_message().id == MSG_ID_STATUS) {
-            Status status = (Status)((comm.received_message().data)[0]);
-            switch (status) {
-                case STATUS_IDLE:   puts("Status: IDLE"); break;
-                case STATUS_MOVING: puts("Status: MOVING"); break;
-                case STATUS_HOMING: puts("Status: HOMING"); break;
-                default:            puts("ERR: Invalid status"); break;
-            }
-        }
-        else {
-            printf("ERROR: Wrong ID: %d\n", comm.received_message().id);
+        switch (status) {
+            case STATUS_IDLE:   puts("Status: IDLE"); break;
+            case STATUS_MOVING: puts("Status: MOVING"); break;
+            case STATUS_HOMING: puts("Status: HOMING"); break;
+            case STATUS_FAULT:  puts("Status: FAULT"); break;
+            default:            puts("ERR: Invalid status"); break;
         }
     }
     else {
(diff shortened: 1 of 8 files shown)


MessageTerminal/MessageTerminal.cxx
```diff
@@ -86,95 +86,42 @@ void move(istringstream& iss)
...
-                case STATUS_IDLE:   puts("Status: IDLE"); break;
-                case STATUS_MOVING: puts("Status: MOVING"); break;
-                case STATUS_HOMING: puts("Status: HOMING"); break;
-                default:            puts("ERR: Invalid status"); break;
-            }
-        }
-        else {
-            printf("ERROR: Wrong ID: %d\n", comm.received_message().id);
-        }
+        printf("Position: (%d, %d)\n", position.x, position.y);
```
chuandudx: Is this position in counts? Would you mind adding a comment :)?

MessageTerminal/MessageTerminal.cxx
```diff
@@ -86,95 +86,42 @@ void move(istringstream& iss)
...
-    if (cmd_name == "home")       return home;
-    if (cmd_name == "move")       return move;
-    if (cmd_name == "stop")       return stop;
-    if (cmd_name == "get_data")   return get_data;
+    if (cmd_name == "ping")           return ping;
+    if (cmd_name == "get_status")     return get_status;
+    if (cmd_name == "home")           return home;
+    if (cmd_name == "move")           return move;
+    if (cmd_name == "stop")           return stop;
+    if (cmd_name == "get_position")   return get_position;
```
chuandudx, line 124: Would it make sense for the `cmd_name`s to be a `typedef enum{}`?

firmware/lib/TestStandCommController/src/TestStandCommController.cxx
```diff
@@ -41,11 +42,15 @@ SerialResult TestStandCommController::status(Status status)
     return this->session.send_message(msg);
 }
 
-SerialResult TestStandCommController::data(uint8_t *data, uint8_t length)
+SerialResult TestStandCommController::position(int32_t x, int32_t y)
```
chuandudx: Should the type be `uint32_t`? are we expecting negative values?

firmware/lib/TestStandCommController/src/TestStandCommController.h
```diff
@@ -18,7 +18,9 @@ class TestStandCommController : public TestStandComm
         SerialResult ping();
         SerialResult log(LogLevel log_level, const char *fmt, ...);
         SerialResult status(Status status);
-        SerialResult data(uint8_t *data, uint8_t length);
+        SerialResult position(int32_t x, int32_t y);
```
chuandudx: Similar comment, position (assuming in counts) would always be a positive number? Since everything is defined wrt the origin (0,0)?

shared/TestStandComm/Messages.h
```diff
@@ -3,33 +3,37 @@
...
 
 // Arduino -> PC Messages
-#define MSG_ID_LOG         0x80
-#define MSG_ID_STATUS      0x81
-#define MSG_ID_DATA        0x82
+#define MSG_ID_LOG              0x80
+#define MSG_ID_STATUS           0x81
+#define MSG_ID_POSITION         0x82
+#define MSG_ID_AXIS_STATE       0x83
+#define MSG_ID_TEMP             0x84
```
chuandudx, line 36: so clean!!

shared_linux/TestStandCommHost/TestStandCommHost.h
```diff
@@ -5,6 +5,11 @@
 
 #include "shared_defs.h"
 
+typedef struct {
+    int32_t x;
+    int32_t y;
+} Position;
+
```
chuandudx, line 12: Same comment as above about `int32_t` vs. `uint32_t`

firmware/lib/TestStandCommController/src/TestStandCommController.cxx
noaht11: Yes, the position is allowed to be negative (I made this change in the motor_refactor PR). There's not really any reason to limit it (it would just create possible wraparound bugs that would be hard to track down)

shared_linux/TestStandCommHost/TestStandCommHost.h
noaht11: See above

firmware/lib/TestStandCommController/src/TestStandCommController.h
noaht11: See above

MessageTerminal/MessageTerminal.cxx
noaht11: Good point, I'll change the variable name and print string

MessageTerminal/MessageTerminal.cxx
noaht11: Enums can only be integers, so that's not an option unfortunately.

firmware/lib/TestStandCommController/src/TestStandCommController.cxx
alexproskurin: Is this position specified in counts?

alexproskurin: I assume all distances in firmware are in counts or steps, right?

noaht11: Yes counts

PR #19, opened 2020-04-04
Restructured project to better scope files and fixed compiler warnings

noaht11:
- Moved firmware library source code into src directory within each library folder to match platformio convention
- Created Gantry.h to hold shared structs / enums for the Gantry library (added as a dependency for MessageTerminal)
  - Axis no longer depends on shared_defs
- Renamed Direction --> AxisDirection
- Changed SerialDevice read function to pass value through pointer and return success boolean to handle compiler warning
- Added casts to log message to handle compiler warning
- Deleted platformio README files


MessageTerminal/Makefile
```diff
@@ -13,13 +13,15 @@ BUILD_DIR = build
 
 LIB_SHARED = ../shared
 LIB_SHARED_LINUX = ../shared_linux
+LIB_FIRMWARE = ../firmware/lib
 
 LIB_TSC = $(LIB_SHARED)/TestStandComm
 LIB_UTIL = $(LIB_SHARED)/Util
 LIB_TSCH = $(LIB_SHARED_LINUX)/TestStandCommHost
 LIB_LSD = $(LIB_SHARED_LINUX)/LinuxSerialDevice
+LIB_GANTRY = $(LIB_FIRMWARE)/Gantry/include
```
chuandudx, line 22: makefile wizard 🥇

firmware/lib/Comm/src/ArduinoSerialDevice.cxx
```diff
@@ -27,9 +27,10 @@ uint32_t ArduinoSerialDevice::ser_available()
     return this->device.available();
 }
 
-uint8_t ArduinoSerialDevice::ser_read()
+bool ArduinoSerialDevice::ser_read(uint8_t *out)
 {
-    return this->device.read();
+    *out = this->device.read();
+    return true;
```
chuandudx, line 33: Does this always return `true`? Should there be an `else` statement

firmware/lib/Gantry/include/Gantry.h
```diff
@@ -0,0 +1,27 @@
...
+ * 
+ * @brief Possible return values for a call to axis_start
+ */
+typedef enum {
+    AXIS_OK,                  //!< Axis movement started OK
+    AXIS_ERR_ALREADY_MOVING,  //!< Axis is already moving
+    AXIS_ERR_LS_HOME,         //!< Trying to move backward while HOME limit switch is pressed
+    AXIS_ERR_LS_FAR,          //!< Trying to move forward while FAR limit switch is pressed
+    AXIS_ERR_INVALID          //!< The parameters resulted in an invalid motion profile
+} AxisResult;
```
chuandudx, line 25: Is there a message for when the axis has reached it's destination successfully?

firmware/lib/Comm/src/ArduinoSerialDevice.cxx
noaht11: Intentionally always returns true. On Arduino there's no way to tell if the read was successful or not, so we have to assume it was

firmware/lib/Gantry/include/Gantry.h
noaht11: There is no message, it's up to MIDAS to monitor the response from `GET_STATUS`.
- `MOVING`: Still moving
- `IDLE`: Reached destination
- `FAULT`: Hit a limit switch before reaching destination

firmware/lib/Comm/src/ArduinoSerialDevice.h
```diff
@@ -15,7 +15,7 @@ class ArduinoSerialDevice: public SerialDevice
         bool ser_connect(uint32_t baud_rate);
```
alexproskurin, line 15: Same as in another PR - maybe remove `ser_` from class methods

noaht11: See reply in the other PR

PR #20, opened 2020-04-06
Improved serial communication robustness

noaht11:
- Moved message id definitions into TestStandMessages.h (separate from the library Messages.h)
- Message data payloads are now specified as structs, so their size and field order is well determined
- Message IDs and data lengths are now checked in TestStandComm
- Implemented AXIS_RESULT message in response to the MOVE command so the errors can be handled on the host side
- Removed acceleration from the MOVE message (it's part of the gantry config now)
- Added explicit (uint8_t *) pointer casts to byte order macros

### DIFF
--- a/MessageTerminal/MessageTerminal.cxx
+++ b/MessageTerminal/MessageTerminal.cxx
@@ -6,6 +6,7 @@
 #include "Gantry.h"
 
 #include "shared_defs.h"
+#include "TestStandMessages.h"
 #include "macros.h"
 
 #include <unistd.h>
@@ -42,19 +43,11 @@ BASIC_CMD(stop);
 
 void move(istringstream& iss)
 {
-    uint32_t accel, hold_vel, dist;
     AxisId axis;
     AxisDirection dir;
+    uint32_t vel_hold, dist;
 
     do {
-        // accel, hold_vel, dist
-        if (!iss.good()) break;
-        iss >> accel;
-        if (!iss.good()) break;
-        iss >> hold_vel;
-        if (!iss.good()) break;
-        iss >> dist;
-
         string word;
 
         // axis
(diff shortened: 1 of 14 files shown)


MessageTerminal/MessageTerminal.cxx
```diff
@@ -6,6 +6,7 @@
 #include "Gantry.h"
 
 #include "shared_defs.h"
+#include "TestStandMessages.h"
```
chuandudx, line 9: snake case? (just recalling from another PR). Hmm, but it seems that other TestStand related headers are camel case too. I must be missing some convention you are using.

MessageTerminal/MessageTerminal.cxx
```diff
@@ -83,7 +90,7 @@ void move(istringstream& iss)
         return;
     } while(0);
 
-    cout << "usage: move <accel> <hold_vel> <dist> <x|y> <pos|neg>" << endl;
+    cout << "usage: move <x|y> <pos|neg> <hold_vel> <dist>" << endl;
```
chuandudx, line 93: good call on remembering to remove `accel` 🥇

MessageTerminal/MessageTerminal.cxx
noaht11: The "convention" is shared_defs is different than everything else so I need to go change it just haven't got to it yet.
Methods and variables are snake_case, classes are PascalCase, files are PascalCase

shared_linux/TestStandCommHost/TestStandCommHost.cxx
```diff
@@ -17,11 +20,9 @@ SerialResult TestStandCommHost::get_status(Status *status_out, uint32_t timeout_
     SerialResult res = this->send_basic_msg(MSG_ID_GET_STATUS);
     if (res != SERIAL_OK) return res;
 
-    res = this->recv_message(timeout_ms);
+    res = this->recv_message(MSG_ID_STATUS, 1, timeout_ms);
```
chuandudx, line 23: What does `1` mean?

noaht11: 1 is the expected data length

PR #21, opened 2020-04-06
Updated feArduino to be compatible with latest serial changes

noaht11:
- Replaced ODB paths and vars with #defines
- Extracted most custom code into ArduinoHelper.cxx (i.e. code that interfaces with our custom libraries)
- Simplified/refactored start_move operations
- Cleaned up error handling / print statements
- Updated gantry mech specifications / calculations for Igus gantry


MidasFrontend/ArduinoHelper.cxx
```diff
@@ -0,0 +1,195 @@
...
+
+#define BAUD_RATE           115200
+#define MSG_RECEIVE_TIMEOUT 5000
+
+LinuxSerialDevice device;
+TestStandCommHost comm(device);
+
+// Gantry Mechanical Properties
+const float pulley_diameter_mm   = 17;
+const float dist_per_rev_mm      = 2 * M_PI * (pulley_diameter_mm / 2);
```
chuandudx, line 23: I'm still unsure whether having integers like `2` would cause truncations in the calculation and then subsequently get casted into a `float` since the datatype is float. Would it be safer to add decimal places. ie `2.0 * M_PI * (pulley_diameter_mm / 2.0)`

MidasFrontend/ArduinoHelper.cxx
```diff
@@ -0,0 +1,195 @@
...
+
+    // Attempt movement
+    bool x_success = attempt_move_axis(AXIS_X, cur_pos_counts.x_counts, dest_mm[AXIS_X], vel_mm_s[AXIS_X]);
+    bool y_success = attempt_move_axis(AXIS_Y, cur_pos_counts.y_counts, dest_mm[AXIS_Y], vel_mm_s[AXIS_Y]);
+
+    // Handle results
+    if (x_success && y_success) {
+        printf("Moving to position (%f mm, %f mm)\n", dest_mm[AXIS_X], dest_mm[AXIS_Y]);
+        printf("Moving with velocity (%f mm/s, %f mm/s)\n", vel_mm_s[AXIS_X], vel_mm_s[AXIS_Y]);
+    }
```
chuandudx, line 184: Should we have a case where we display if x succeeded but y didn't, and vice versa?

MidasFrontend/ArduinoHelper.cxx
noaht11: If at least one of the operands is a float then the result will be a float, so we're safe here. That being said, doesn't hurt to add the `.0` so I can go ahead and add it

firmware/src/main.cpp
```diff
@@ -54,15 +54,15 @@ const mPMTTestStandConfig conf = {
...
-        .vel_home_a             = 10000,  // holding velocity for homing A [steps / s]
-        .vel_home_b             = 10000,  // holding velocity for homing B [steps / s]
-        .accel_home_a           = 5000,   // acceleration for homing A [steps / s^2]
-        .accel_home_b           = 5000    // acceleration for homing B [steps / s^2]
+        .accel                  = 10,  // acceleration for all motion [steps / s^2]
+        .vel_start              = 1,   // starting velocity for all motion [steps / s]
+        .vel_home_a             = 75,  // holding velocity for homing A [steps / s]
+        .vel_home_b             = 75,  // holding velocity for homing B [steps / s]
+        .accel_home_a           = 10,  // acceleration for homing A [steps / s^2]
+        .accel_home_b           = 10   // acceleration for homing B [steps / s^2]
```
chuandudx, line 65: How were these values chosen and why are they so much smaller than the previous `steps/s` values?

MidasFrontend/ArduinoHelper.cxx
noaht11: I'm saying if either of them failed then the "move" as a whole failed, and we stop both axes. I don't think we want to do half moves where one axis moves, but the other doesn't. Still working on nicer error reporting.

firmware/src/main.cpp
noaht11: They were chosen fairly arbitrarily. They're for the new gantry so there's no gear reduction and we need way lower speeds than before

MidasFrontend/ArduinoHelper.cxx
chuandudx: makes sense :)

MidasFrontend/ArduinoHelper.cxx
alexproskurin: Just an architectural? question - what's the general rule for using `#define  CONSTANT 7` vs `const int = 7;`

MidasFrontend/ArduinoHelper.cxx
```diff
@@ -0,0 +1,195 @@
...
+// Gantry Mechanical Properties
+const float pulley_diameter_mm   = 17;
+const float dist_per_rev_mm      = 2 * M_PI * (pulley_diameter_mm / 2);
+const float mm_cts_ratio         = dist_per_rev_mm / ENCODER_COUNTS_PER_REV;
+const float mm_steps_ratio       = dist_per_rev_mm / MOTOR_STEPS_PER_REV;
+
+// Operational Bounds
+const float gantry_x_min_mm      = 0.0;
+const float gantry_x_max_mm      = 1200.0; // max rail is 1219 mm
+const float gantry_y_min_mm      = 0.0;
```
alexproskurin, line 30: nit: Would just add comments saying `mm per second`

alexproskurin: Referring to the one below haha

MidasFrontend/ArduinoHelper.cxx
```diff
@@ -0,0 +1,195 @@
...
+    if (dest_mm[AXIS_Y] < gantry_y_min_mm || dest_mm[AXIS_Y] > gantry_y_max_mm) {
+        cm_msg(
+            MERROR,
+            "validate_move_params",
+            "Destination on y-axis should be between %f mm and %f mm inclusive.",
+            gantry_y_min_mm, gantry_y_max_mm);
+        return false;
+    }
+
+    if (vel_mm_s[AXIS_X] < gantry_vel_min_mm_s || vel_mm_s[AXIS_X] > gantry_vel_max_mm_s
```
alexproskurin, line 59: Instead of erroring out when a user specifies velocity > max, it might be more user friendly to give a WARNING, but execute command with hardware max velocity.

alexproskurin: Same for min velocity

MidasFrontend/ArduinoHelper.cxx
```diff
@@ -0,0 +1,195 @@
...
+static int32_t mm_to_cts(float val_mm) {
+    return round(val_mm / mm_cts_ratio);
+}
+
+static uint32_t mm_to_steps(float val_mm) {
+    return round(val_mm / mm_steps_ratio);
+}
+
+static AxisDirection get_direction(int32_t displacement) {
+    return (displacement < 0 ? AXIS_DIR_NEGATIVE : AXIS_DIR_POSITIVE);
```
alexproskurin, line 81: had to look up what `?` did, pretty neat!

alexproskurin: Will abuse it now HAHAHA

MidasFrontend/ArduinoHelper.cxx
```diff
@@ -0,0 +1,195 @@
...
+}
+
+/*****************************************************************************/
+/*                             PUBLIC FUNCTIONS                              */
+/*****************************************************************************/
+
+/**
+ * @brief Establishes a serial connection to the Arduino
+ * 
+ * This includes opening the serial device as well as waiting to receive a
```
alexproskurin, line 128: would add argument description

MidasFrontend/ArduinoHelper.cxx
```diff
@@ -0,0 +1,195 @@
...
+ * ping message from the Arduino to validate that it is running
+ * 
+ * @return true if the connection was successfully established, otherwise false
+ */
+bool arduino_connect(char *device_file)
```
alexproskurin, line 135: is timeout defined in device file? or does this not timeout?

MidasFrontend/ArduinoHelper.cxx
```diff
@@ -0,0 +1,195 @@
...
+}
+
+/**
+ * @brief Attempts to command the Arduino to move to the provided destination at the
+ *        provided velocity
+ * 
+ * @param dest_mm   Pointer to two floats (the absolute x and y coordinates in mm)
+ * @param vel_mm_s  Pointer to two floats (the x and y velocities in mm/s)
+ */
+void arduino_attempt_move(float *dest_mm, float *vel_mm_s)
```
alexproskurin: Would remove attempt, because in my mind every function attempts to do things haha, would just change it to `arduino_move`

MidasFrontend/feArduino.cxx
```diff
@@ -149,129 +138,37 @@ HNDLE handleMove;
...
-  path += "/Settings";
+  INT status;
 
   // Get the destination position (absolute distance)
-  string destpath = path + "/Destination";
   float destination[2] = {0,0};
   int size_dest = sizeof(destination);
-  if (db_get_value(hDB, 0, destpath.c_str(), &destination, &size_dest, TID_FLOAT, TRUE) != DB_SUCCESS) {
-      // TODO error message
+  if ((status = db_get_value(hDB, 0, ODB_VAR_DESTINATION, &destination, &size_dest, TID_FLOAT, TRUE)) != DB_SUCCESS) {
```
alexproskurin: finding kinda hard to read, maybe would be nicer if it was split on function calls that get assigned to variables with and then check condition based on those variables?

MidasFrontend/feArduino.cxx
```diff
@@ -149,129 +138,37 @@ HNDLE handleMove;
...
-  if (comm.move(accel_default_cts,
-                user_vel_x_cts,
-                user_rel_dist_x_cts,
-                AXIS_X,
-                user_x_dir)) {
-    printf("Move x-direction OK\n");
-  } else {
-    printf("Error sending move command in x-direction\n");
+  if ((status = db_get_value(hDB, 0, ODB_VAR_VELOCITY, &velocity, &size_vel, TID_FLOAT, TRUE)) != DB_SUCCESS) {
+    cm_msg(MERROR, "start_move", "Failed to retrieve velocity (%s) from ODB. Error: %d", ODB_VAR_VELOCITY, status);
```
alexproskurin, line 163: Same as above

MidasFrontend/ArduinoHelper.cxx
noaht11: Oh no, what have I done!

MidasFrontend/ArduinoHelper.cxx
noaht11: I think we're better off erroring out. I don't think we should ever execute any movements under parameters that we think the user intended. What if they typed an extra zero on a slow velocity so the velocity goes out of bounds and now we're running at max speed when they only wanted to run slowly. I think there's lots of scenarios where this could go wrong.

MidasFrontend/ArduinoHelper.cxx
noaht11: This is a fun one.
`#define MAGIC 7` (in this case) comes from me forgetting that I'm writing C++ and not C. In C, `const int magic = 7;` is not actually a "constant expression" which means there's a whole bunch of scenarios where you can't use it like a constant (e.g. case labels, array dimensions). Essentially, `const` variables in C are not truly constant, so the general practice is to use `#define`s or `enum`s instead which are truly constant.
In C++, `const` variables are truly constant (well sometimes you have to declare them `constexpr`), so there's actually very few scenarios if any where you would use a `#define`.

So given that MIDAS is now C++ ish sorta kinda, I should probably swap out my defines for consts.

MidasFrontend/ArduinoHelper.cxx
noaht11: Oh whoops yeah, the argument wasn't there originally, forgot to add a description when I added it

MidasFrontend/ArduinoHelper.cxx
noaht11: There is no timeout for this. This is the first thing that happens when you start the program so the idea is to just sit there and wait until the Arduino pings us.

MidasFrontend/ArduinoHelper.cxx
noaht11: There's so many things that could go wrong with a move that it felt extra "attempty", so I added it to the name, but yeah I agree, every function is an attempt, so will remove

MidasFrontend/feArduino.cxx
noaht11: Sure, this is a fairly common paradigm that's kind of up for debate as to whether or not it's bad practice. I don't feel too strongly about it so I can change it.

MidasFrontend/ArduinoHelper.cxx
alexproskurin: Ok, cool. This is good to know!

noaht11: These constants ended up moving in the next PR so I will switch them from defines later

PR #22, opened 2020-04-08
Implement MIDAS event updating (and various fixes)

noaht11:
- Implemented status and position updates in read_arduino_state and fixed linux platform_millis()

 **Implemented temperature reporting and refactored so strict aliasing rules are no longer violated**
- Implemented temperature messages in serial application layer
- Replaced casts of uint8_t * to struct ____ * with memcpy calls
- Replaced ntohl and htonl macros with inline functions

**Fixed serial communication and temperature reporting**
Serial Communication
- Implemented link_check and echo messages to be able to verify that all bytes can be sent across the serial link
- Added link_check call to initial connection setup in ArduinoHelper
- Disabled iexten option in LinuxSerialDevice which was swallowing 0x16 and 0x0F bytes
- Baud rates are now defined by an enum
- Fixed baud rate setting through termios by using the actual termios defines for baud rates
- Moved ping definition into the parent class (TestStandComm)
- Moved send_buf into parent class (TestStandComm)

feArduino
- Deleted unused code in read_arduino_state
- Fixed data type of temperature bank pointer (now a double* instead of float*)

MessageTerminal
- Fixed MessageTerminal Makefile (added TempMeasure library dependency)
- Added link_check command to run the link check
- Added get_temp command to MessageTerminal

Misc
- Moved Debug.h back into library folder so libraries can access it


MessageTerminal/MessageTerminal.cxx
```diff
@@ -190,6 +214,9 @@ int main(int argc, char *argv[])
             device.ser_disconnect();
             if (!connect_to_arduino()) return 1;
```
chuandudx, line 215: Can we make `1` more descriptive

MidasFrontend/ArduinoHelper.cxx
```diff
@@ -1,22 +1,20 @@
...
-#define MSG_RECEIVE_TIMEOUT 5000
-
-LinuxSerialDevice device;
-TestStandCommHost comm(device);
+/*****************************************************************************/
+/*                                 CONSTANTS                                 */
+/*****************************************************************************/
 
 // Gantry Mechanical Properties
 const float pulley_diameter_mm   = 17;
```
chuandudx, line 20: `17.0` for consistency

MidasFrontend/ArduinoHelper.cxx
```diff
@@ -73,6 +92,10 @@ static int32_t mm_to_cts(float val_mm) {
     return round(val_mm / mm_cts_ratio);
 }
 
+static float cts_to_mm(int32_t val_cts) {
```
chuandudx, line 95: should we make this `uint32_t`? I do remember from a previous PR you mentioned position is generalized so it may be positive or negative, but encoder counts are always positive right? Or does the system interpret negative counts as going in the opposite direction?

MidasFrontend/feArduino.cxx
```diff
@@ -34,6 +34,11 @@ Arduino frontend for mPMT test stand.
 #define ODB_VAR_DESTINATION  ODB_PATH_SETTINGS "/Destination"
 #define ODB_VAR_VELOCITY     ODB_PATH_SETTINGS "/Velocity"
 
+// Note: these must be exactly 4 letters long
+#define ODB_BANK_GANTRY      "GANT"
+#define ODB_BANK_STATUS      "STAT"
```
chuandudx, line 39: Can you remind me what STAT tells us? Was it a flag that indicates the system is moving?

MidasFrontend/feArduino.cxx
```diff
@@ -337,64 +342,44 @@ int Nloop, Ncount;
...
+    double *pddata_temp;
+    bk_create(pevent, ODB_BANK_TEMP, TID_DOUBLE, (void**)&pddata_temp);
+    *pddata_temp++ = temp_data.temp_ambient;
+    *pddata_temp++ = temp_data.temp_motor_x;
+    *pddata_temp++ = temp_data.temp_motor_y;
+    *pddata_temp++ = temp_data.temp_mpmt;
+    *pddata_temp++ = temp_data.temp_optical;
+    bk_close(pevent, pddata_temp);
+  }
+
```
chuandudx, line 384: Do we store the actual user-input position in case there is a slight mismatch between distance traveled vs. desired? Wondering if that would go in the same bank as well?

shared_linux/TestStandCommHost/TestStandCommHost.cxx
```diff
@@ -71,8 +68,32 @@ SerialResult TestStandCommHost::get_position(PositionMsgData *position_out, uint
...
+
+    // Copy message data into output struct
+    TempMsgData msg_data;
+    memcpy(&msg_data, this->received_message().data, sizeof(TempMsgData));
+    // Fixup byte order and re-scale
+    temp_out->temp_ambient = (double)ntohl(msg_data.temp_ambient) / temp_data_scaler;
+    temp_out->temp_motor_x = (double)ntohl(msg_data.temp_motor_x) / temp_data_scaler;
+    temp_out->temp_motor_y = (double)ntohl(msg_data.temp_motor_y) / temp_data_scaler;
+    temp_out->temp_mpmt    = (double)ntohl(msg_data.temp_mpmt) / temp_data_scaler;
+    temp_out->temp_optical = (double)ntohl(msg_data.temp_optical) / temp_data_scaler;
```
chuandudx, line 96: What is the `1E6` scaler?

firmware/lib/Comm/src/TestStandCommController.cxx
```diff
@@ -58,6 +50,24 @@ SerialResult TestStandCommController::position(int32_t x_counts, int32_t y_count
...
 }
 
+SerialResult TestStandCommController::temp(TempData *temp_data)
+{
+    TempMsgData data = {
+        .temp_ambient = htonl(round(temp_data->temp_ambient * temp_data_scaler)),
+        .temp_motor_x = htonl(round(temp_data->temp_motor_x * temp_data_scaler)),
+        .temp_motor_y = htonl(round(temp_data->temp_motor_y * temp_data_scaler)),
+        .temp_mpmt    = htonl(round(temp_data->temp_mpmt    * temp_data_scaler)),
+        .temp_optical = htonl(round(temp_data->temp_optical * temp_data_scaler))
```
chuandudx, line 60: How many decimal places do we report for temperature data?

firmware/lib/Comm/src/TestStandCommController.h
```diff
@@ -17,12 +19,11 @@ class TestStandCommController : public TestStandComm
     public:
         TestStandCommController(SerialDevice& device);
 
-        SerialResult ping();
         SerialResult log(LogLevel log_level, const char *fmt, ...);
         SerialResult status(Status status);
         SerialResult position(int32_t x_counts, int32_t y_counts);
         SerialResult axis_state(/*TODO*/);
```
chuandudx, line 25: Will there be another PR with this implementation?

firmware/src/main.cpp
```diff
@@ -84,6 +87,7 @@ void setup()
 
 void loop()
 {
+    // PROFILE("Test Stand", 1000, test_stand.execute());
     test_stand.execute();
 
     // Blink an LED
```
chuandudx, line 93: Is this functionality to visually get an idea if serial communication is working? Just a note that with the shield installed, we won't be able to see the LED blink :(

shared/TestStandMessages.h
```diff
@@ -61,4 +57,12 @@ typedef struct {
     int32_t y_counts;
 } __attribute__((__packed__)) PositionMsgData;
 
+typedef struct {
+    int32_t temp_ambient;
+    int32_t temp_motor_x;
+    int32_t temp_motor_y;
+    int32_t temp_mpmt;
+    int32_t temp_optical;
```
chuandudx, line 65: can these be `uint32_t`

shared/shared_defs.h
```diff
@@ -1,6 +1,10 @@
 #ifndef SHARED_DEFS_H
 #define SHARED_DEFS_H
 
+#define SERIAL_BAUD_RATE       BAUD_115200
+
+#define MSG_RECEIVE_TIMEOUT_MS 250
+
 #define ENCODER_COUNTS_PER_REV 500
 #define MOTOR_STEPS_PER_REV    800
```
chuandudx, line 9: Has the microsteps variable been removed? Is there another place where we may set it.

MidasFrontend/ArduinoHelper.cxx
noaht11: Encoder counts can be negative, otherwise things can break with weird wrap-around conditions

firmware/lib/Comm/src/TestStandCommController.h
noaht11: Yup

firmware/src/main.cpp
noaht11: This doesn't tell us anything about serial communication necessarily, just lets me know the program hasn't stalled or crashed. You can sorta kinda see it with the shield if you look at it sideways right?

shared/TestStandMessages.h
noaht11: Could the temperature not be negative?

MidasFrontend/feArduino.cxx
noaht11: It currently only stores the result of `GET_STATUS` (either moving, homing, idle etc.). Could potentially throw other stuff in there as well in the future. Might throw the limit switch statuses in.

firmware/lib/Comm/src/TestStandCommController.cxx
noaht11: Currently 6 (completely arbitrarily chosen), that's what the `temp_data_scaler = 1E6` multiplier is.
We read the number as a double, say `99.12345678`
Multiple by `1E6` and round: `99.12345678 * 1E6 = 99123456.78 --> 99123457`
Transfer the value as an integer: `99123457`
Divide by `1E6` on the host side: `99123457 / 1E6 = 99.123457`

Transferring doubles is a pain (because of endianness, pointer aliasing, possible different binary representations etc.) so this seemed like a reasonable approach given the range of numbers and precision we're expecting.

shared_linux/TestStandCommHost/TestStandCommHost.cxx
noaht11: See my reply up above where you asked about the number of decimal places

shared/TestStandMessages.h
chuandudx: You're right, the thermistor range is actually between -55 to 150. I forgot that it could be negative since I was only imagining above 0 temperatures for the mPMTs.

